PR #82, opened 2021-09-11
fix(GaussianState): GBS sampling is biased

Kolarovszki:
We tried to be too smart.

The previous calculation introduced a bias, since it divided the
currently calculated probability by the previously calculated one to
acquire conditional probabilities. This is wrong, since the previous
probability was conditioned on a different heterodyne measurement,
corresponding to a different quantum state.

It hasn't been noticed with the hypothesis test. The issue was noticed
by observing that the set of conditional probabilities didn't always sum
up to one.

The calculation was rewritten to omit the previous probability.
Additionally, the normalization and the previos samples from the
occupation number factorial in the denominator can be omitted, since it
is the same for every "probabilites". Omitting these factors, the
variables got renamed to "weight(s)" from "probabilit(y|ies)".

To make the hypothesis test more reliable, the shots got increased to
5000 from only 1000.

Additionally, the `Q` matrix has been refactored into the `Q_matrix`
property.

### DIFF
--- a/piquasso/_backends/gaussian/state.py
+++ b/piquasso/_backends/gaussian/state.py
@@ -471,6 +471,10 @@ class GaussianState(State):
             ]
         ) + np.identity(2 * self.d)
 
+    @property
+    def Q_matrix(self):
+        return (self.complex_covariance + np.identity(2 * len(self))) / 2
+
     def rotated(self, phi: float) -> "GaussianState":
         r"""Returns the copy of the current state, rotated by angle `phi`.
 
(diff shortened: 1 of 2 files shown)


piquasso/_backends/gaussian/state.py
```diff
@@ -1073,60 +1068,29 @@ def _get_particle_number_choice_and_probability(
 
         d = len(self)
 
-        Q = (self.complex_covariance + np.identity(2 * d)) / 2
-
-        Qinv = np.linalg.inv(Q)
-
-        B = - Qinv[d:, :d]
+        B = - np.linalg.inv(self.Q_matrix)[d:, :d]
```
icnots, line 1071: This refactor is very efficient (eliminating all the calculations with A matrix) but is not so clear from the code. Does it appear in the documentation?

Kolarovszki: Yeah, it is in the docstring!

icnots (approved):

PR #88, opened 2021-09-26
refact(GaussianState): Threshold simplification

Kolarovszki:
The `ThresholdMeasurement` using the torontonian is simplified, since
the GBS algorithm doesn't use the same `GaussianState._perform_sampling`
method anymore, since e9322042e857f8ae9293f5b87f216bb167afe55d.


kareem1925 (commented):
it works :+1: great catch btw

PR #131, opened 2021-12-16
Feature(quadrature variance):`fock/pure/state`

kareem1925:
- The method `quadratures_mean` has been changed to `quadratures_mean_variance`
  to return the expectation and the variance of the quadratures for a specific mode.

- The test case has been modified to check for the returned variance values.

This feature is needed for the pennylane-piquasso plugin.

### DIFF
--- a/docs/tutorials/getting-started.ipynb
+++ b/docs/tutorials/getting-started.ipynb
@@ -2,6 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "a66a4f7d",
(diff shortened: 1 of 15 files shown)


Kolarovszki:
Maybe `quadratures_mean_and_variance` would be a better name, but otherwise it seems good!

Kolarovszki:
Maybe the docstring fixes in this commit can be put in a separate commit next time.

Kolarovszki (approved):

PR #145, opened 2022-01-10
feat(fock): Wigner function

Kolarovszki:
The Wigner function in the `BaseFockState` class has been copyed from
QuTiP.

The Wigner function from the `BaseFockState` is compared to the result
from `GaussianState` in a test.


boldar99 (approved):
LGTM

PR #147, opened 2022-01-10
Restructure code generation

boldar99:
A new function, `as_code` has been implemented in a new file, `piquasso.api.utils` that is responsible for transforming an existing program, simulator, and shots into code.

To implement this function, the following updates has been made:
- An `__eq__` method has implemented in the `Config` class
- Both `Config`, `Simulator`, and `Program` now implement `CodeMixin` so that they have `_as_code` methods.

Tests have been written to all the implemented methods.

### DIFF
--- a/piquasso/_backends/fock/general/calculations.py
+++ b/piquasso/_backends/fock/general/calculations.py
@@ -34,7 +34,10 @@ def passive_linear(state: FockState, instruction: Instruction, shots: int) -> Re
     operator: np.ndarray = instruction._all_params["passive_block"]
 
     fock_operator = state._space.get_passive_fock_operator(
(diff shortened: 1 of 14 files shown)


kareem1925 (approved):

tests/api/program/test_code_generation.py
```diff
@@ -75,15 +65,10 @@ def test_code_generation_with_empty_program():
     with pq.Program() as program:
         pass
 
-    code = program.as_code()
+    code = program._as_code()
```
Kolarovszki, line 68: I would rewrite these tests to use the public `as_code` function, or simply delete these tests, since it tests a private method.

PR #156, opened 2022-01-20
perf(fock/pure): Avoid embedding of active gates

Kolarovszki:
Some gate implementation embedded the matrix into the larger Fock space,
then multiplied the state vector, which is quite inefficient. This is
rewritten to calculate the new vector coefficients directly, instead.

This will be done for the `general` Fock simulation and similar for the
passive gates as well.


kareem1925 (approved):
great!

PR #170, opened 2022-03-31
feat(api): Rename `errors` to `exceptions`

Kolarovszki:
The `errors` module got renamed to `exceptions`, since the latter is
more appropriate.


icnots (approved):
lgtm

PR #175, opened 2022-04-07
FIX(clements): Handling 0 matrix elements

Kolarovszki:
During test writing for the Clements decomposition under
`piquasso/decompositions/clements.py`, an issue has been encountered.
Namely, there is a possible zero division error during the
decomposition: one needs to handle the zero division case separately in
the `Clements` class.


icnots (approved):
Great work.

PR #184, opened 2022-04-25
fix(FockSimulator): Wrong auxiliary modes in calculation

Kolarovszki:
In the calculation functions for the `FockSimulator`, the auxiliary
modes were passed incorrectly to `embed_matrix`, since the auxiliary
modes were determined from the whole `instruction.modes` instead of just
the loop variable `mode`. This has been fixed.

Equivalence tests have been written accordingly.


kareem1925 (approved):
great catch @Kolarovszki :+1:

PR #188, opened 2022-04-26
fix(Kerr): Kerr gate scaling

Kolarovszki:
Some tests were not entirely thorough on the scaling of the `Kerr` and
the `CubicPhase` gates. Additional tests were written, and it turned out
that the `Kerr` gate is not working.

The issue was, that the `kerr` calculation in `FockSimulator` and
`PureFockSimulator` was written in a way that only the first mode was
considered for a multimode calculation. This has been fixed by rewriting
the logic to consider multiple modes.

Several tests were written for testing the scalability of the `Kerr` and
`CubicPhase` gates. Furthermore, the `test_kerr_equivalency` has not
been testing the equivalence between the `FockSimulator` and
`PureFockSimulator`, therefore it has been rewritten to do that.

Additional tests have been written testing the code generation with the
`Kerr` and `CubicPhase` gates.

### DIFF
--- a/piquasso/_backends/fock/general/calculations.py
+++ b/piquasso/_backends/fock/general/calculations.py
@@ -134,16 +134,18 @@ def annihilate(state: FockState, instruction: Instruction, shots: int) -> Result
 
 
 def kerr(state: FockState, instruction: Instruction, shots: int) -> Result:
-    mode = instruction.modes[0]
-    xi = instruction._all_params["xi"]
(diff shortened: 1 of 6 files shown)


kareem1925:
It would be great if another we test code generation with this fix. Otherwise, nicely done!

kareem1925 (approved):
Works perfectly and the generated code is valid. 👌

PR #191, opened 2022-08-17
Feat Homodyne sampling

kareem1925:
* Added homodyne sampling in the Fock simulator and Pure Fock simulator
* Added required tests
* Added hermite function in `functions.py`

### DIFF
--- a/piquasso/_backends/fock/general/calculations.py
+++ b/piquasso/_backends/fock/general/calculations.py
@@ -13,15 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 from typing import Tuple, Mapping
+from itertools import product
 
 import random
 import numpy as np
+from scipy.special import factorial
 
 from .state import FockState
 
 from piquasso.api.instruction import Instruction
 from piquasso.api.result import Result
+from piquasso._math.functions import hermite
 
 
 def vacuum(state: FockState, instruction: Instruction, shots: int) -> Result:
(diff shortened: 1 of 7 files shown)


piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
+            modes=instruction.modes,
+            d=reduced_state._space.d,
+            permanent_function=reduced_state._config.permanent_function,
+        )
+        reduced_state._density_matrix = (
+            fock_operator
+            @ reduced_state._density_matrix
+            @ fock_operator.conjugate().transpose()
+        )
+    resolution = 100000
```
Kolarovszki, line 332: This is not a good approach. You should try to do this without approximation, since there is an exact distribution one can calculate.

Also, one can compare in a test the resulting distribution with e.g. the distribution generated by the GaussianSimulator.

piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
     state._state_vector[index] = coefficient
+
+
+def measure_homodyne(
+    state: PureFockState, instruction: Instruction, shots: int
+) -> Result:
+    modes = instruction.modes
+    phi = instruction._all_params["phi"]
+    reduced_state = state.reduced(modes=modes)
+    scaled_hbar = 1 / reduced_state._config.hbar
```
Kolarovszki, line 317: `hbar_inv`, rather, but I think this variable is quite redundant.

piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
+
+    hermite_matrix = np.zeros((cutoff, cutoff, resolution))
+    for n, m in product(range(cutoff), repeat=2):
+        hermite_matrix[n][m] = (
+            1.0
+            / np.sqrt(2**n * factorial(n) * 2**m * factorial(m))
+            * hvalues[n]
+            * hvalues[m]
+        )
+    hermite_values = np.expand_dims(reduced_state.density_matrix, -1) * np.expand_dims(
```
Kolarovszki, line 351: This backend is for calculations with pure state. Couldn't we make this calculation more efficient knowing that we have a state vector?

piquasso/_backends/fock/general/calculations.py
```diff
@@ -287,3 +291,69 @@ def _add_occupation_number_basis(
     dual_index = state._space.index(bra)
 
     state._density_matrix[index, dual_index] = coefficient
+
+
+def measure_homodyne(state: FockState, instruction: Instruction, shots: int) -> Result:
```
Kolarovszki, line 296: This function is almost exactly the same as the one for the pure one. Try not to repeat yourself.

piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
+            permanent_function=reduced_state._config.permanent_function,
+        )
+        reduced_state._density_matrix = (
+            fock_operator
+            @ reduced_state._density_matrix
+            @ fock_operator.conjugate().transpose()
+        )
+    resolution = 100000
+    cutoff = reduced_state._config.cutoff
+    quadrature_range = np.linspace(-cutoff, cutoff, resolution)
```
Kolarovszki, line 334: Why is this the range for calculating the distribution? What is the interpretation of the cutoff in phase space?

piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
+    )
+
+    pdf_rho /= np.sum(pdf_rho)
+
+    pdf_rho[np.abs(pdf_rho) < 1e-11] = 0
+
+    homodyne_result = []
+    for _ in range(shots):
+        homodyne_result.append(
+            quadrature_range[np.where(np.random.multinomial(1, pdf_rho) == 1)[0]]
```
Kolarovszki, line 373: I don't quite get how the multinomial distribution comes into play here.

Kolarovszki: Also, this produces a single value, right? But if you make a measurement on multiple modes, you should acquire multiple values, right? See `GaussianSimulator`.

piquasso/_backends/fock/pure/calculations.py
```diff
@@ -303,3 +306,71 @@ def _add_occupation_number_basis(  # type: ignore
...
+        )
+    # The remaining state is not collapsed but the reduced density matrix
+    return Result(state=reduced_state, samples=homodyne_result)
```
Kolarovszki, line 376: You should pass the evolved state here.

piquasso/_backends/fock/pure/simulator.py
```diff
@@ -105,5 +107,6 @@ class PureFockSimulator(Simulator):
         gates.Squeezing2: linear,
         gates.GaussianTransform: linear,
         measurements.ParticleNumberMeasurement: particle_number_measurement,
+        measurements.HomodyneMeasurement: measure_homodyne,
```
Kolarovszki, line 110: the function should be named `homodyne_measurement` as the other measurements.

tests/backends/fock/general/test_measurements.py
```diff
@@ -205,3 +205,29 @@ def test_measure_particle_number_with_multiple_shots():
...
+    simulator = pq.FockSimulator(d=2, config=pq.Config(cutoff=10))
+
+    result_position = simulator.execute(position, shots)
+
+    result_momentum = simulator.execute(momentum, shots)
+
+    assert len(result_momentum.samples) == shots
+    assert len(result_position.samples) == shots
+    assert np.isclose(np.array(result_position.samples).mean(), 1, atol=0.1)
+    assert np.isclose(np.array(result_momentum.samples).mean(), 0, atol=0.1)
```
Kolarovszki, line 233: This is a very unhelpful and possibly flaky test. Test the distribution instead and write a script which compares the samples from the `PureFockSimulator` and `GaussianSimulator` via a hypothesis test.

tests/backends/fock/pure/test_measurements.py
```diff
@@ -157,3 +157,29 @@ def test_measure_particle_number_with_multiple_shots():
...
+    simulator = pq.PureFockSimulator(d=2, config=pq.Config(cutoff=10))
+
+    result_position = simulator.execute(position, shots)
+
+    result_momentum = simulator.execute(momentum, shots)
+
+    assert len(result_momentum.samples) == shots
+    assert len(result_position.samples) == shots
+    assert np.isclose(np.array(result_position.samples).mean(), 1, atol=0.1)
+    assert np.isclose(np.array(result_momentum.samples).mean(), 0, atol=0.1)
```
Kolarovszki, line 185: This is also a very unhelpful test, but if you have 2 tests which are basically the same except the simulator, I suggest using `pytest.mark.parametrize`.

Kolarovszki (changes requested):
Also, write more commit message, try to write down how you came up with the calculation, refer to sources, and use conventional commits: https://www.conventionalcommits.org/en/v1.0.0/

PR #194, opened 2022-08-31
Tf fock gates

kareem1925:
Non gaussian gates support for TF

- Added logm and expm methods since TF doesn't support their gradients
- Modified Kerr, CrossKerr, and CubicPhase gates to support TensorFlow
computational style and to enable calculating the gradients.
- Tests were added to check for the sanity of the functions.


piquasso/_math/fock.py
```diff
@@ -21,7 +21,8 @@
 from operator import add
 
 from scipy.special import factorial, comb
-from scipy.linalg import expm
+
+# from scipy.linalg import expm
```
Kolarovszki, line 25: Delete this!

Kolarovszki (changes requested):
Please name your commit as `feat(fock.pure): Tensorflow for non-linear gates` or similar.

Also, make sure to have equivalence tests between the tensorflow and the original simulator.

tests/backends/tensorflow/test_gradient.py
```diff
@@ -153,6 +153,50 @@ def test_Squeezing_mean_photon_number_gradient_1_mode():
...
+
+        mean = state.mean_photon_number()
+
+    gradient = tape.gradient(mean, [gamma])
+
+    assert np.isclose(mean, 0.029567)
+    assert np.isclose(gradient, 0.291416)
+
+
+def test_Kerr_gates_mean_photon_number_gradient():
```
Kolarovszki, line 177: Can you add a test case when the Kerr gate produces a non-zero gradient?

Kolarovszki: Also, test for the CrossKerr gate is missing.

tests/backends/tensorflow/test_gradient.py
```diff
@@ -153,6 +153,50 @@ def test_Squeezing_mean_photon_number_gradient_1_mode():
...
+            pq.Q(0) | pq.CubicPhase(gamma=gamma)
+
+        state = simulator.execute(program).state
+
+        mean = state.mean_photon_number()
+
+    gradient = tape.gradient(mean, [gamma])
+
+    assert np.isclose(mean, 0.029567)
+    assert np.isclose(gradient, 0.291416)
```
Kolarovszki, line 174: I think for one mode these could be easily calculated explicitely from the evolution of the ladder operators.

PR #195, opened 2022-09-05
fix(sampling): Photon detection with uniform loss

Kolarovszki:
**Problem**

The lossy boson sampling with uniform losses produced only zero-array
samples.

**Solution**

It turned out, that the class
`GeneralizedCliffordsUniformLossesSimulationStrategy` has an optional
parameter `transmissivity` and Piquasso is supposed to specify that
parameter.

Some tests have been written testing the singular values of the
interferometer after specifying loss.

The documentation of `pq.Loss` was also unclear about the purpose of
`transmissivity`, it has also been updated.

### DIFF
--- a/piquasso/_backends/sampling/calculations.py
+++ b/piquasso/_backends/sampling/calculations.py
@@ -169,6 +169,9 @@ def _get_sampling_simulation_strategy(
     _, singular_values, _ = np.linalg.svd(state.interferometer)
 
     if np.all(np.isclose(singular_values, singular_values[0])):
-        return GeneralizedCliffordsUniformLossesSimulationStrategy(permanent_calculator)
+        uniform_transmission_probability = singular_values[0] ** 2
+        return GeneralizedCliffordsUniformLossesSimulationStrategy(
+            permanent_calculator, uniform_transmission_probability
+        )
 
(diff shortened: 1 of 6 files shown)


piquasso/_backends/sampling/calculations.py
```diff
@@ -169,6 +169,9 @@ def _get_sampling_simulation_strategy(
     _, singular_values, _ = np.linalg.svd(state.interferometer)
 
     if np.all(np.isclose(singular_values, singular_values[0])):
-        return GeneralizedCliffordsUniformLossesSimulationStrategy(permanent_calculator)
+        uniform_loss_probability = singular_values[0] ** 2
```
Tomev, line 172: This actually is the uniform probability of TRANSMITTANCE not of the loss.

Tomev (approved):
I only have one minor issue with the variable name. Aside from that I'm good.

piquasso/_backends/sampling/calculations.py
Kolarovszki: done

Kolarovszki:
> I only have one minor issue with the variable name. Aside from that I'm good.

That is something that we need to discuss in a meeting.

PR #247, opened 2023-11-08
feat(gaussian): `GaussianState.purify`

Kolarovszki:
A method called `GaussianState.purify` has been implemented, which returns with the purification of the `GaussianState` instance.

### DIFF
--- a/piquasso/_backends/gaussian/state.py
+++ b/piquasso/_backends/gaussian/state.py
@@ -15,6 +15,8 @@
 
 from typing import Optional, Tuple, List
 
+from scipy.linalg import block_diag
+
 import numpy as np
 
 from piquasso.api.config import Config
@@ -30,6 +32,8 @@ from piquasso._math.linalg import (
 from piquasso._math.symplectic import symplectic_form
 from piquasso._math.combinatorics import get_occupation_numbers
 from piquasso._math.transformations import from_xxpp_to_xpxp_transformation_matrix
+
+from piquasso._math.decompositions import williamson
 
 from .probabilities import (
     DensityMatrixCalculation,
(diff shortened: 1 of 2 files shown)


piquasso/_backends/gaussian/state.py
```diff
@@ -861,3 +865,62 @@ def fock_probabilities(self) -> np.ndarray:
         return calculation.get_particle_number_detection_probabilities(
             get_occupation_numbers(d=self.d, cutoff=self._config.cutoff)
         )
+
+    def is_pure(self) -> bool:
```
boldar99, line 869: Do you use the formula from the paper mentioned in ln.887?
If so, why do you use Delta instead of Delta^(-1)?

Kolarovszki: Thank you for the comment, and sorry for the late reply!

Which formula do you refer to?

boldar99: (3.19) of https://arxiv.org/pdf/quant-ph/9912067.pdf

Kolarovszki: Oh, I see. I just used the following argument:

The relation can easily seen by noting that a covariance matrix $\Sigma$ corresponds to a pure state iff $\Sigma = S S^T$ for some $S \in \mathrm{Sp}(2d, \mathbb{R})$ (here, the covariance matrix is already normalized, i.e., $\hbar = 1$). By the symplectic relation, this is equivalent to
```math
(\Delta \Sigma)^2 = \Delta S \underbrace{S^T \Delta S}_{=\Delta} S^T = \Delta \underbrace{S \Delta S^T}_{=\Delta} = \Delta^2 = - I.
```

Comparing this with Eq. (3.19), we can substitute $\hbar=1$, because we normalized the covariance matrix beforehand. Therefore, $\Delta^{-1} = - \Delta$, hence we can write $(\Delta^{-1} \Sigma)^2 = (\Delta \Sigma)^2$, and it does not matter if we take the inverse or not. The factor $1/4$ appears on the RHS because their covariance matrix definition is half of ours.

boldar99: Nice

boldar99 (approved):

PR #307, opened 2024-04-14
script: Train a CZ gate with imperfect postselection

czaba7:
A training script has been implemented to train a six mode KLM CZ gate on dual-rail encoded qubits with imperfect postselection to increase the ouput fidelity

### DIFF
--- a/scripts/train_cz_gate_with_imperfect_detectors.py
+++ b/scripts/train_cz_gate_with_imperfect_detectors.py
@@ -1,0 +1,267 @@
+#
+# Copyright 2021-2024 Budapest Quantum Computing Group
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


scripts/train_cz_gate.py
```diff
@@ -0,0 +1,273 @@
...
+        ]
+    )
+
+    interferometer = get_interferometer_from_weights(
+        weights, d, calculator, config.complex_dtype
+    )
+
+    program = pq.Program(
+        instructions=[
+            *preparation.instructions,
```
Kolarovszki, line 131: I think it is not necessary to create `preparation` here. You can just paste the `pq.StateVector` instructions here.

scripts/train_cz_gate.py
```diff
@@ -0,0 +1,273 @@
...
+        ]
+    )
+
+    simulator = pq.PureFockSimulator(d=4, config=config, calculator=calculator)
+    expected_state = simulator.execute(expected_program).state
+
+    return expected_state.density_matrix
+
+
+def get_ideal_weights():
```
Kolarovszki, line 64: `get_initial_weights` maybe? This could be confusing because these weights are not ideal when imperfect detectors are used.

Kolarovszki (changes requested):
Maybe it could be renamed `script/train_cz_gate_with_imperfect_detectors.py`?

Kolarovszki (approved):

PR #321, opened 2024-07-18
perf(fock): jitted `calculate_interferometer_helper_indices`

VHenrik007:
Several functions have been rewritten so that they can be compiled using Numba JIT. This is a partial milestone and more functions are planned to be rewritten in this manner.

Specifically, the following changes occurred:
- The `comb` function is now part of the `combinatorics.py` module, and an additional `arr_comb` is implemented accepting and returning arrays. 
- `partitions` is now also jitted alongside `nb_combinations`, equivalent to `itertools.combinations` which is used in many places, hence the different name with the `nb_` prefix. 
- The function `get_fock_space_basis` has also a new implementation with a `nb_` prefix denoting the jitted version. The distinction is necessary due to the extensive use of LRU caching at other parts of the code.
- The function `get_index_in_fock_subspace_array` now can be compiled using Numba JIT.
- And finally `calculate_interferometer_helper_indices` is also now jitted with Numba. 

Notes:
- For the special case of the parametrization of `partitions` the shape `(1,0)` is specifically chosen so that it can be stacked in `get_fock_space_basis` correctly. 
- The function header for `cutoff_cardinality_array` got changed so that it does not force keyword arguments. This seems to be a bug in Numba related to keyword arguments when they are called in another function that is also JIT compiled. (see issue [here](https://github.com/numba/numba/issues/9185))
- The function `calculate_interferometer_helper_indices` now returns tuples instead of dictionaries (due to Numba) and therefore a temporary dictionary is added between function calls until the entire `get_interferometer_on_fock_space` function gets Numba JIT implementation as well.

### DIFF
--- a/piquasso/_math/fock.py
+++ b/piquasso/_math/fock.py
@@ -20,6 +20,7 @@ import numpy as np
 import numba as nb
 
 from scipy.special import comb
+from piquasso._math.combinatorics import comb as nb_comb
 
 from piquasso.api.config import Config
 from piquasso._math.combinatorics import partitions
@@ -47,8 +48,14 @@ def cutoff_cardinality(*, cutoff: int, d: int) -> int:
     return comb(d + cutoff - 1, cutoff - 1, exact=True)
 
 
-def cutoff_cardinality_array(*, cutoff, d):
-    return np.round(comb(d + cutoff - 1, d)).astype(int)
+@nb.njit
+def cutoff_cardinality_array(cutoff, d):
+    ret = np.empty(cutoff.shape, dtype=np.int32)
+
+    for i in range(len(cutoff)):
+        ret[i] = nb_comb(d + cutoff[i] - 1, d)
+
+    return ret
 
 
 def symmetric_subspace_cardinality_array(*, d: int, n: int) -> int:
--- a/piquasso/_math/fock.py
+++ b/piquasso/_math/fock.py
@@ -17,6 +17,7 @@ import functools
 from typing import Tuple
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


piquasso/_math/fock.py
```diff
@@ -46,17 +48,34 @@ def cutoff_cardinality(*, cutoff: int, d: int) -> int:
...
+        ret[i] = nb_comb(d + cutoff[i] - 1, d)
+
+    return ret
 
 
 def symmetric_subspace_cardinality_array(*, d: int, n: int) -> int:
     return np.round(comb(d + n - 1, n)).astype(int)
 
 
-@functools.lru_cache(maxsize=None)
```
Kolarovszki, line 57: I think you don't have to delete this cache. The `cache=True` in `njit` is to avoid recompilation.

Kolarovszki (changes requested):

piquasso/_math/fock.py
VHenrik007: Unfortunately, this function is called inside another `njit`-compiled function which results in an error since it cannot compile `@functools.lru_cache`:
`Cannot determine Numba type of <class 'functools._lru_cache_wrapper'>`

Kolarovszki: One can reintroduce this by creating a different function which is cached. This function is called repeatedly during a calculation with mostly the same arguments. Even if it got faster (3-fold speedup in this case), it would be still nice to have it cached in cases when it is not used inside a jitted function.

Kolarovszki (approved):

PR #336, opened 2024-09-18
refact: Rename `backend` to `simulator`

Kolarovszki:
The term 'backend' is an archaic naming stemming from older versions of Piquasso. The naming got stuck, because the original `piquasso._backends` package contained not only `Simulator` classes, but also `State` classes.

To make things simpler, this naming is deleted altogether, and rewritten to 'simulator', a more appropriate naming IMHO.

We might also consider moving the `State` classes into a different `piquasso._states` package.


antalszava (approved):
Looks good to me :ok_hand:

PR #345, opened 2024-10-07
refact: Rename `Calculator` to `Connector`

Kolarovszki:
The class which encapsulated the underlying numerical/ML framework was named `Calculator` in the past. However, this naming does not describe well its purpose, therefore it got renamed to `Connector`.


antalszava (approved):
Looks good to me :+1:

PR #375, opened 2024-10-28
chore: Add Python 3.13 support

Kolarovszki:


Kolarovszki:
Will be reopened when Numba supports 3.13.

PR #429, opened 2025-04-16
perf: pfaffian C++ implementation

Mou1z:
Wrote a C++ implementation of the Pfaffian. Improved the performance using Parlett-Reid Algorithm instead of using the algorithm based on Schur decomposition.

Added `build.bat` for building the project on Windows.


src/pfaffian.cpp
```diff
@@ -0,0 +1,91 @@
+#include <cmath>
+
+#include "matrix.hpp"
+#include "pfaffian.hpp"
+
+template <typename TScalar>
+TScalar pfaffian_cpp(Matrix<TScalar> &matrix_in) {
```
Kolarovszki, line 32: Can you also include a docstring, citing the implemented algorithm (https://arxiv.org/abs/1102.3440)?

src/pfaffian.cpp
```diff
@@ -0,0 +1,116 @@
...
+                matrix_in[i_k1] = matrix_in[i_kp];
+                matrix_in[i_kp] = tmp;
+            }
+
+            result *= -1;
+        }
+
+        TScalar element = matrix_in[(k * n) + k + 1];
+
+        if(element != 0) {
```
Kolarovszki, line 82: This comparison might not be very reliable, what if `element` is really close to `0.0`? I guess it would cause numerical accuracy issues for near-singular matrices. Can you check what happens with a singular and a near-singular matrix?

piquasso/_math/pfaffian.cpp
```diff
@@ -0,0 +1,48 @@
...
+    TScalar result = pfaffian_cpp(native_matrix);
+
+
+    return create_numpy_scalar(result);
+}
+
+const char* pfaffian_docstring = R""""(
+Calculates the pfaffian of a matrix.
+
+Note:
```
Kolarovszki, line 40: I don't think you need a `Note:` here. Maybe you can write a bit more about this function, but it is optional.

build.bat
```diff
@@ -0,0 +1,23 @@
+@echo off
```
Kolarovszki, line 1: At the beginning of the file, can you explain what is it for? Also, can you mention this in `README.md`?

Kolarovszki (changes requested):
All in all, it is a very nice work, thank you! I have just a few suggestions.

build.bat
Kolarovszki: No, I meant the whole file. Somewhere I think it needs to be stated that this is a helper for building locally on windows.

Mou1z: Oh okay. I will do it now.

src/pfaffian.cpp
Mou1z: I added a new test case `test_pfaffian_6_by_6_skew_symmetric_small_random` testing the function on matrices with very small values. The test cases pass. I will have to check for singular / near-singular matrices.

src/pfaffian.cpp
Mou1z: Yes, I included it in the docstring now.

build.bat
Mou1z: I updated the `README.md` and `build.bat` files, you can review them.

src/pfaffian.cpp
Kolarovszki: (We discussed that we will deal with potential numerical instabilities later.)

Kolarovszki (approved):

Issue #441, opened 2025-05-21
Add Python 3.13 support


Kolarovszki:
### What should be added?

Piquasso does not support Python 3.13 yet. We tried to do it in https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/413, but we ran into the issue that Tensorflow does not support Python 3.13. This is not necessarily a problem, but our tests (and hence, the CI/CD pipeline) depend on Tensorflow due to our optional `tensorflow` plugin.

To solve this, one needs to somehow "turn off" the Tensorflow-related tests when testing with Python 3.13. The implementation might need to use `pytest` markers, probably custom marker for labeling tests that use Tensorflow or maybe [`skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1), but these might not be ideal. The hacker is invited to implement the most elegant approach. **It is also important to only "turn off" Tensorflow-related tests, and let other tests run in Python 3.13.**

Code in the source folder `piquasso` should not be touched ideally.

EDIT:

Maybe it is beneficial to use `pytest.importorskip`. This would trivially work if the tests would only import tensorflow, but, unfortunately they also instantiate `TensorflowConnector` classes as well, sometimes at the top of the test files. A step to solving the issue is to create fixtures returning with these, i.e.,
```python
import sys

@pytest.fixture
def tf():
    if sys.version_info.minor >= 13:
        return pytest.importorskip("tensorflow")

    import tensorflow

    return tensorflow


@pytest.fixture
def TensorflowConnector(tf):
    return pq.TensorflowConnector
```
and to use these fixtures in tests instead of importing and instantiating `TensorflowConnector` at the top of the file. This might be problematic because `TensorflowConnector` is also used in parametric tests.

Hope-Alemayehu:
Hi @Kolarovszki,
I’ve worked on this issue as part of the UnitaryHack hackathon and submitted a PR. When you have a moment, could you please review it? Thank you!

Kolarovszki:
> Hi [@Kolarovszki](https://github.com/Kolarovszki), I’ve worked on this issue as part of the UnitaryHack hackathon and submitted a PR. When you have a moment, could you please review it? Thank you!

Hi @Hope-Alemayehu! Are you still working on this issue?

Kolarovszki:
> Any updates when do we get a Tensorflow support for Python 3.13?

[...]

PR #442, opened 2025-05-28
 Add Python 3.13 Support by Skipping TensorFlow Tests #441 

Hope-Alemayehu:
This PR fixes [#441](https://github.com/piquasso/piquasso/issues/441) by skipping TensorFlow-related tests when running with Python ≥ 3.13, since TensorFlow doesn't yet support this version.

- Added pytest_ignore_collect hook to exclude specific test files known to use TensorFlow.

- No changes were made to the main piquasso source code.

- Ensures tests and CI pass cleanly under Python 3.13 without requiring TensorFlow.

#UnitaryHack

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import os
+import sys
 import pytest
 import numpy as np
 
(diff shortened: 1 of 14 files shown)


tests/conftest.py
```diff
@@ -30,6 +31,29 @@
...
+        
+        # Skip test files that are known to import TensorFlow
+        tf_related_files = [
+            "test_simulator_equivalence.py",
+            "test_tf_function.py",
+            "test_clements.py",
+            "test_gates.py",
+            "test_decompositions.py",
+        ]
+        return any(filename in path_str for filename in tf_related_files)
```
Kolarovszki, line 53: So there is an issue with this, which is that it skips entire files, skipping tests also that are not using tensorflow.

For example, in the test `tests/_simulators/fock/test_gates.py::test_squeezing_probabilities`. We would still like to have it run for all the possible `SimulatorClass`es except for the ones that use `tensorflow`.

My first idea here is to the tensorflow import in a fixture, and when using Python 3.13, disable tests using this fixture. I don't know if this can be directly achieved in pytest, but I'm sure there is a solution.

Kolarovszki (changes requested):

PR #448, opened 2025-05-29
Fix tensorflow import compatibility #441

Hope-Alemayehu:
Add Python 3.13 support by conditionally skipping TensorFlow-related tests when TensorFlow isn't available or compatible.

Details

- Skips TensorFlow tests on Python 3.13+ using logic in conftest.py
- Added clearer error messages for missing TensorFlow/JAX
- Cleaned up and fixed TensorFlow-related test fixtures

Testing
- Verified TensorFlow tests are skipped on unsupported Python versions
- Confirmed all other tests pass on supported environments
- Non-TensorFlow failures (e.g., in test_hafnian.py and simulator equivalence tests) remain unrelated to this change

#UnitaryHack

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,7 @@
 [pytest]
 markers =
     monkey: marks tests with stochastic inputs.
+    tensorflow: mark test as requiring TensorFlow
 
 python_functions = test_* *_benchmark *_script
 python_files = test_*.py *_benchmark.py *_script.py
--- a/tests/_math/test_decompositions.py
+++ b/tests/_math/test_decompositions.py
@@ -32,7 +32,23 @@ from piquasso._simulators.connectors import (
 )
 
 
-@pytest.mark.parametrize("connector", [NumpyConnector(), TensorflowConnector()])
+# Create a list of connectors to test
+CONNECTORS = [NumpyConnector()]
+
+try:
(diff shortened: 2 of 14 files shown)


Kolarovszki (changes requested):
I don't understand this patch. It seems that multiple strategies are used, and maybe this is the best solution, but I don't see this yet. I expected something much more simple to work.

Ideally, the test writer should not be required to write, e.g.,
```
    if sys.version_info >= (3, 13):
        pytest.skip("TensorFlow not supported on Python 3.13+")
```
in a test, as you did in `test_clements_decomposition_with_TensorflowConnector_from_weights` -- it is just too much work to handle these at such a low level. Also, it could happen that `tensorflow` will be supported with Python 3.13 soon, and in that case, somebody has to go through every one of these tests to check if tensorflow is not disabled. 

I think putting `tensorflow` and `jax` in fixtures are good, maybe one can use a simple `importorskip` if we are on Python 3.13. But then there are still conditional imports and `importorskip` at the beginning of files, and I don't understand why. Of course, `TensorflowConnector` should not be instantiated when there is no `tensorflow`, but `TensorflowConnector` can also be put in a fixture, right? Also, there is a marker for marking tensorflow tests, but it is not clear why this is needed besides all of the code you implemented.

I think we should talk in a meeting or something to discuss this.

PR #413, opened 2025-02-22
chore: Add Python 3.13 support

Kolarovszki:

PR #450, opened 2025-06-01
Feat/python3.13

Hazarre:
One solution for https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues/441.


Kolarovszki (changes requested):
Yeah, this is definitely a solution!

However, the tests are pretty mixed, some files only partially use Tensorflow, as we often need the same code to be executed via Tensorflow, JAX and Numpy, and I don't think we should skip these. Somebody else ran into this problem a while ago in https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/448.

The other solution would be to -- instead of importing Tensorflow in the tests -- put Tensorflow import in a fixture, and disable tests using this fixture on Python 3.13. This might be bit of a challenge, because there are also parametric tests using Tensorflow, but I'm sure there is an elegant solution for this in pytest (see discussion under https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/448).

Hazarre:
> Hi @Hazarre!
> 
> The problem is that there are many tests that have to run `Connector`-agnostically, meaning that has to run successfully for multiple `*Connector`s. Therefore, these are implemented in a parametrized manner. In fact, I want to extend JAX and Tensorflow support in the future, so there will be more tests like this. Consequently, separating tensorflow-related tests would yield major code duplication I'd like to avoid.

That is going to be non-trivial since tests will need info about Connectors dependency on tf. 

Might be useful reference.
https://github.com/PennyLaneAI/pennylane/blob/master/tests/conftest.py

Kolarovszki:
> > Hi @Hazarre!
> > The problem is that there are many tests that have to run `Connector`-agnostically, meaning that has to run successfully for multiple `*Connector`s. Therefore, these are implemented in a parametrized manner. In fact, I want to extend JAX and Tensorflow support in the future, so there will be more tests like this. Consequently, separating tensorflow-related tests would yield major code duplication I'd like to avoid.
> 
> That is going to be non-trivial since tests will need info about Connectors dependency on tf.
> 
> Might be useful reference. https://github.com/PennyLaneAI/pennylane/blob/master/tests/conftest.py

Hi @Hazarre,

Of course, this is a non-trivial task -- if it would be trivial, it would have been done already  in my original PR (https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/413). And yes, an example for this could be Pennylane, although we have slightly different challenges in Piquasso.

If you investigate a bit further, you will find out that Piquasso itself only uses Tensorflow through `TensorflowConnector`, which has to be specified through the `*Simulator` classes. Skipping tests requiring this connector would roughly solve one part of the problem.

PR #453, opened 2025-06-02
Add `get_marginal_fock_probabilities` method to Fock and Gaussian

### LINKED ISSUE #436: Marginal particle detection probabilities
### What should be added?

**Problem**

In Piquasso, one can typically calculate the particle number detection probabilities of quantum states by using the property [`fock_probabilities`](https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/5524a2e8b558f3a4b9e4de48ef09bb924f50fdfc/piquasso/_simulators/fock/state.py#L82). If one would like to calculate the marginal probabilities corresponding to a subsystem specified by a tuple of integers called `modes`, one can use the [`reduced(modes)`](https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/5524a2e8b558f3a4b9e4de48ef09bb924f50fdfc/piquasso/_simulators/fock/state.py#L67) method to acquire a quantum state restricted to the subsystem defined by `modes`, and then get the marginal probabilities by using `fock_probabilities`. 

However, it might be convenient to have a function which calculates the marginal probabilities.

**Potential solution**

Implement a `get_marginal_fock_probabilities` method for [`GaussianState`](https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/5524a2e8b558f3a4b9e4de48ef09bb924f50fdfc/piquasso/_simulators/gaussian/state.py#L45), [`PureFockState`](https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/5524a2e8b558f3a4b9e4de48ef09bb924f50fdfc/piquasso/_simulators/fock/pure/state.py#L36) and [`FockState`](https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/5524a2e8b558f3a4b9e4de48ef09bb924f50fdfc/piquasso/_simulators/fock/general/state.py#L33). This could be a common method which issues a call to the `reduced` method and then uses the `fock_probabilities` method of the resulting state object to provide marginal probabilities.

Note: the main input argument to `get_marginal_fock_probabilities` could be the list of modes just as for `reduced` - a list of integers signifying the indices of modes to reduce the state to before obtaining the marginal probabilities.

**Testing**

The written code should be accompanied by unit tests written in `tests/_simulators`.

q-inho:
Resolves #436 
- Implemented a convenience method to obtain marginal particle number probabilities. The method calls `reduced` internally and returns `fock_probabilities` for the selected subsystem
- Added unit tests confirming that the new method gives the same results as reducing the state explicitly, covering Fock, PureFock, and Gaussian simulators

### DIFF
--- a/tests/_simulators/fock/pure/test_state.py
+++ b/tests/_simulators/fock/pure/test_state.py
@@ -105,7 +105,7 @@ def test_PureFockState_get_marginal_fock_probabilities():
     state = simulator.execute(program).state
 
     modes = (1,)
-    expected_probabilities = state.reduced(modes).fock_probabilities
+    expected_probabilities = np.array([0.5, 0.25, 0.25, 0.0])
     actual_probabilities = state.get_marginal_fock_probabilities(modes)
 
     assert np.allclose(actual_probabilities, expected_probabilities)
--- a/tests/_simulators/fock/pure/test_state.py
+++ b/tests/_simulators/fock/pure/test_state.py
@@ -94,6 +94,23 @@ def test_PureFockState_fock_probabilities_map():
         )
 
 
+def test_PureFockState_get_marginal_fock_probabilities():
+    with pq.Program() as program:
+        pq.Q() | pq.StateVector([0, 1]) / 2
+
+        pq.Q() | pq.StateVector([0, 2]) / 2
+        pq.Q() | pq.StateVector([2, 0]) / np.sqrt(2)
+
+    simulator = pq.PureFockSimulator(d=2)
+    state = simulator.execute(program).state
+
+    modes = (1,)
+    expected_probabilities = state.reduced(modes).fock_probabilities
+    actual_probabilities = state.get_marginal_fock_probabilities(modes)
+
+    assert np.allclose(actual_probabilities, expected_probabilities)
+
+
 def test_PureFockState_quadratures_mean_variance():
     with pq.Program() as program:
         pq.Q() | pq.Vacuum()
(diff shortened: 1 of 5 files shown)


q-inho:
Realigning the format is done!

tests/_simulators/fock/pure/test_state.py
```diff
@@ -94,6 +94,23 @@ def test_PureFockState_fock_probabilities_map():
         )
 
 
+def test_PureFockState_get_marginal_fock_probabilities():
```
Kolarovszki, line 97: Using `PureFockState`, we can test the values explicitely, not relying on `reduced` or `fock_probabilities`, and this would be nice. You have also shown examples for this in your comment using `FockState` and `PureFockState`, demonstrating that the marginal probabilities are `[0.5  0.25 0.25 0.  ]`. These are good to include to verify that the calculation makes sense (for me, at least, but I guess in general). Can you include these as tests? For `GaussianState`, it is not necessary I think, because this probability distribution can be a bit difficult to calculate explicitely.

Kolarovszki (changes requested):
This indeed looks good! I have just a minor suggestion!

tests/_simulators/fock/pure/test_state.py
q-inho: Updated to `expected_probabilities = np.array([0.5, 0.25, 0.25, 0.0])` for both `FockState` and `PureFockState`

Kolarovszki (approved):
Looks good to me! Thank you for your contribution!

PR #466, opened 2025-06-10
perf(_math): Permanent C++ implementation

0xSooki:
Reimplementing the permanent in C++.

This takes inspiration from the original PiquassoBoost implementation found in https://github.com/Budapest-Quantum-Computing-Group/piquassoboost/blob/main/piquassoboost/sampling/source/BBFGPermanentCalculatorRepeated.hpp which is also reimplemented in https://github.com/0xSooki/permanent-boost.

The formula used is from Eq. (8) in https://arxiv.org/pdf/2309.07027.pdf.


src/permanent.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // determine the concurrency of the calculation
+    unsigned int n_threads = std::thread::hardware_concurrency();
+    int64_t concurrency = static_cast<int64_t>(n_threads) * 4;
+    concurrency = concurrency < idx_max ? concurrency : static_cast<int64_t>(idx_max);
+
+    std::vector<TComplex> thread_results(static_cast<unsigned int>(concurrency), TComplex(0.0, 0.0));
+
+// See: https://stackoverflow.com/a/8447025
+#if defined(_OPENMP)
+#pragma omp parallel for num_threads(concurrency)
```
Kolarovszki, line 153: There is a new error here on Windows, similar to the previous comment. Is there a reason to use `int64_t` for `work_batch`, can't we just use `uint64_t` instead, and for all the variables derived from this?

Kolarovszki (approved):
Thank you!

PR #486, opened 2025-07-03
Add `Result.get_counts`

antalszava:
Adds the `Result.get_counts` method that depends on the `Result.samples` attribute.

### DIFF
--- a/piquasso/api/result.py
+++ b/piquasso/api/result.py
@@ -41,9 +41,15 @@ class Result:
         return f"Result(samples={self.samples}, state={self.state})"
 
     def get_counts(self):
-        if any(not isinstance(sample, (int, np.integer)) for sample_tuple in self.samples for sample in sample_tuple):
-            raise NotImplementedError("The get_counts method only supports measurements that provide integer samples."\
-            "(like particle number measurements).")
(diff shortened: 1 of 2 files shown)


antalszava:
The way I see it, the implemented `get_counts` method is simply a convenience for the user, and as such, I personally wouldn't alter its logic. It simply does what it's there to do: from the samples the simulator generates, we count how many times we obtain each sample. The original motivation is to obtain the counts of the number of states when performing PNR measurements.

Even if floats are in the output samples, I don't think they should be manipulated in any way to get counts (perhaps maybe if due to numerical inaccuracies we o
btain samples that are different only up to some small epsilon, which we could create a tolerance value for). Specifically, if the user performs measurements that yield floats and calls `get_counts`, they should then receive the counts of each sample, which may also be a float. Implementing binning is a complete different type of logic and don't think it's something that should necessarily come within the scope of this PR - it could be an option to have that as a feature, but I think it should even be in a separate method or maybe added as a flag to `get_counts`, but I don't think it's necessary atm.

What do you think?

antalszava:
@Kolarovszki thank you for the suggestions, they've been addressed 🙏

PR #493, opened 2025-07-28
Allow multiple measurements in the same program

antalszava:


tests/_simulators/sampling/test_measurements.py
```diff
@@ -502,3 +502,109 @@ def test_post_select_random_unitary():
...
+        expected_state = expected_simulator.execute(expected_program).state
+
+        assert state == expected_state
+
+    def test_reindexing_for_measurements(self):
+
+        with pq.Program() as program:
+            pq.Q() | pq.StateVector([1, 1, 1, 0, 0])
+            pq.Q() | pq.PostSelectPhotons(postselect_modes=[0, 1], photon_counts=[1, 1])
+            pq.Q(2, 4) | pq.ParticleNumberMeasurement()
```
antalszava, line 594: Hi @Kolarovszki, I'm curious about your thoughts on how this and the following example (``test_measuring_inactive_raises``) work.

The idea would be that if some modes have been measured, they are no longer active (active modes are tracked under the hood by the simulator - an implementation detail that is perhaps less important here for the user-facing part of the feature). Users can index modes with "absolute" indices, that is, for ``d`` number of modes being simulated in the ``Program``, one always indexes modes with indices ``0``, ``1``, ... ``d`` when constructing instructions in the ``Program``, regardless of which modes are active or not. However, if the user attempts to act on an instruction in a mode that is not active, an error is raised (as in the next test case). If not all modes are active in the simulation (because some have already been measured), then under the hood, we remap mode indices so that it makes sense for internal computations, but the user doesn't need to pay attention to which modes are active when providing the indices (except for the fact that inactive modes cannot be acted on with instructions).

How does this user interface look?

PR #498, opened 2025-08-14
Validate input occupation numbers and cutoff value for `pq.StateVector`

### LINKED ISSUE #449: Validate input occupation numbers  and cutoff value when using Fock simulator with `pq.StateVector`
### What should be added?

**Context**

The input to the `pq.StateVector` instruction implicitly determines occupation numbers which also implicitly determine the size of the subspace required to simulate the input state vector, for example, with `pq.PureFockSimulator`. Fock state simulations operate with a user-specified cutoff, and the previously mentioned implicit size definition of the subspace means that the user has to define a good enough cutoff for the simulation to be accommodated. Otherwise, if the cutoff is too small, then the subspace is also too small, and the simulator cannot accommodate the state vector definition. In practice, this results in a non-informative error message in Piquasso because there is a mismatch of a "small" array being indexed with a "big" index.

**Code to reproduce**

In practice, the above-mentioned scenario in Piquasso behaves in a way where an error is raised due to a mismatch in a NumPy array size and the index we're attempting to use.

Consider the following code snippet:
```py
import piquasso as pq

d = 6

cutoff = 4
connector = pq.NumpyConnector()
config = pq.Config(cutoff=cutoff)

state_00 = [0, 1, 0, 1]
ancilla_state = [1, 1]

program = pq.Program(
    instructions=[
        pq.StateVector(state_00 + ancilla_state, coefficient=1)
    ]
)

simulator = pq.PureFockSimulator(d=d, config=config, connector=connector)
simulator.execute(program).state
```
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[35], line 19
     12 program = pq.Program(
     13     instructions=[
     14         pq.StateVector(state_00 + ancilla_state, coefficient=1)
     15     ]
     16 )
     18 simulator = pq.PureFockSimulator(d=d, config=config, connector=connector)
---> 19 simulator.execute(program).state

File ~/software/piquasso/piquasso/api/simulator.py:273, in Simulator.execute(self, program, shots, initial_state)
    252 """Executes the specified program.
    253 
    254 Args:
   (...)
    268         if any measurement is specified in `program`.
    269 """
    271 instructions: List[Instruction] = program.instructions
--> 273 return self.execute_instructions(
    274     instructions, initial_state=initial_state, shots=shots
    275 )

File ~/software/piquasso/piquasso/api/simulator.py:245, in Simulator.execute_instructions(self, instructions, initial_state, shots)
    241     calculation = self._get_calculation(instruction)
    243     instruction = self._maybe_postprocess_batch_instruction(instruction)
--> 245     result = calculation(result.state, instruction, shots)
    247 return result

File ~/software/piquasso/piquasso/_simulators/fock/pure/calculations/__init__.py:415, in state_vector_instruction(state, instruction, shots)
    412 def state_vector_instruction(
    413     state: PureFockState, instruction: Instruction, shots: int
    414 ) -> Result:
--> 415     _add_occupation_number_basis(
    416         state=state,
    417         **instruction._all_params,
    418         modes=instruction.modes,
    419     )
    421     return Result(state=state)

File ~/software/piquasso/piquasso/_simulators/fock/pure/calculations/__init__.py:439, in _add_occupation_number_basis(state, coefficient, occupation_numbers, modes)
    435     occupation_numbers = new_occupation_numbers
    437 index = get_index_in_fock_space(occupation_numbers)
--> 439 state.state_vector = state._connector.assign(state.state_vector, index, coefficient)

File ~/software/piquasso/piquasso/_simulators/connectors/numpy_/connector.py:67, in NumpyConnector.assign(self, array, index, value)
     66 def assign(self, array, index, value):
---> 67     array[index] = value
     69     return array

IndexError: index 169 is out of bounds for axis 0 with size 84
```
The practical reason for the error is that we have `cutoff=4` defined with an underlying `occupation_numbers` array of `array([0, 1, 0, 1, 1, 1])`. Once the cutoff is increased to 5 with this specific state vector input, no error is raised.

**Problem**

The main problem with the above scenario is that the user is not informed what the exact issue is with the inputs that they have provided.

**Suggested solution**

The solution is to add a validation check to ensure that the cutoff is big enough when using the state vector instruction so that no such indexing or related error is raised. With an informative error message, the user can be instructed to increase the cutoff value (or consider another type of simulation with a different occupation number array).

q-inho:
## Summary
- Guarded **bosonic Fock `StateVector`** preparation so that any set of occupation numbers exceeding the configured cutoff now raises `InvalidState` and reports the minimum required cutoff for both direct inputs and `fock_amplitude_map` entries.
- Mirrored cutoff validation for fermionic simulations so every **fermionic `StateVector`** including those built from amplitude maps respects the cutoff limit.
- Added regression tests across bosonic, and fermionic modules that confirm `InvalidState`
- Resolve #449 

## Demo
```python
import piquasso as pq

d = 6

cutoff = 4
connector = pq.NumpyConnector()
config = pq.Config(cutoff=cutoff)

state_00 = [0, 1, 0, 1]
ancilla_state = [1, 1]

program = pq.Program(
    instructions=[
        pq.StateVector(state_00 + ancilla_state, coefficient=1)
    ]
)

simulator = pq.PureFockSimulator(d=d, config=config, connector=connector)
simulator.execute(program).state
```
This demo code now gives
```
InvalidState: The occupation numbers '(0, 1, 0, 1, 1, 1)' require a cutoff of at least '5', but the provided cutoff is '4': instruction=StateVector(occupation_numbers=(0, 1, 0, 1, 1, 1), coefficient=1, modes=(0, 1, 2, 3, 4, 5))
```

### DIFF
--- a/piquasso/_math/validations.py
+++ b/piquasso/_math/validations.py
@@ -55,7 +55,7 @@ def are_modes_consecutive(modes: Tuple[int, ...]) -> bool:
 
 
 def validate_occupation_numbers(
-    occupation_numbers: Iterable, d: int, cutoff: int
+    occupation_numbers: Iterable, d: int, cutoff: int, context: str = ""
 ) -> None:
     """Validate occupation numbers for Fock state preparation.
 
@@ -63,25 +63,33 @@ def validate_occupation_numbers(
         occupation_numbers: Sequence of occupation numbers to validate.
         d: The expected number of modes.
         cutoff: The cutoff dimension for the Fock space.
+        context: Optional message appended to raised errors.
 
     Raises:
         InvalidState: If the length of ``occupation_numbers`` does not match
             ``d`` or if the total particle number requires a larger cutoff.
     """
 
-    occupation_numbers = np.array(occupation_numbers)
+    original_occupation_numbers = tuple(occupation_numbers)
+    occupation_numbers = np.array(original_occupation_numbers)
 
     if len(occupation_numbers) != d:
-        raise InvalidState(
-            f"The occupation numbers '{tuple(occupation_numbers.tolist())}' are "
+        message = (
+            f"The occupation numbers '{original_occupation_numbers}' are "
             f"not well-defined on '{d}' modes."
         )
+        if context:
+            message += context
+        raise InvalidState(message)
 
     total = int(np.sum(occupation_numbers))
     if total >= cutoff:
         required_cutoff = total + 1
-        raise InvalidState(
-            f"The occupation numbers '{tuple(occupation_numbers.tolist())}' require "
+        message = (
+            f"The occupation numbers '{original_occupation_numbers}' require "
             f"a cutoff of at least '{required_cutoff}', but the provided cutoff is "
             f"'{cutoff}'."
         )
+        if context:
+            message += context
+        raise InvalidState(message)
--- a/piquasso/_math/validations.py
+++ b/piquasso/_math/validations.py
@@ -17,6 +17,8 @@ from typing import Iterable, Union, Tuple
 
 import numpy as np
 
+from piquasso.api.exceptions import InvalidState
+
 
 def is_natural(number: Union[int, float]) -> bool:
     return bool(np.isclose(number % 1, 0.0) and round(number) >= 0)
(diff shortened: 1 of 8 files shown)


Kolarovszki (changes requested):
I think this is great! However, to avoid repeating code, I think it would be better to create a function for validating occupation numbers. More concretely, this could be a function called `validate_occupation_numbers` which takes `occupation_numbers`, `d` and `cutoff` as an argument, and raises an error if the `occupation_numbers` is invalid. Or, any similar pattern would be good here which avoids code repetition.

Kolarovszki (approved):

PR #517, opened 2025-10-05
feat(tests): Distributing CI test runs across multiple CPUs using `pytest-xdist`

antalszava:
The test execution is distributed by `pytest-xdist` with the `-n auto` flag.

Documentation: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,7 +22,8 @@ extras =
     py{38,39,310,311,312}: jax,dev,tensorflow,matplotlib,dask
     py313: jax,dev,matplotlib,dask
 deps =
-    pytest-cov,pytest-xdist
+    pytest-cov
+    pytest-xdist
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -23,21 +23,22 @@ extras =
     py313: jax,dev,matplotlib,dask
 deps =
     pytest-cov
+    pytest-xdist
```
Kolarovszki, line 26: But then shouldn't this be listed in the `pyproject.toml` instead?

antalszava: I'm not sure I see significance to moving it there, but I can definitely do so: developers can decide if they want to use this locally or not. The main change is to use it on the CI.

Kolarovszki (approved):

PR #519, opened 2025-10-08
feat!: Branching logic for mid-circuit measurements

Kolarovszki:
An improved logic for handling multiple measurements is added. This replaces
the original `_execute_multi_measurement_instructions`, which is executed when
multiple measurements are detected, and ran the entire circuit independently
`shots` many times. Instead, the measurement outcomes are binned and stored
with their frequency among all samples in a `Branch` object, along with the
corresponding post-measurement quantum state.

This means, that a "branching" is performed at a measurement, and the simulator
needs to update the list branches (instead of a single state) when applying
each instruction. By storing the frequencies of the branches, no two branch has
the same outcome, and hence, no redundant calculations are performed, yielding
a speedup over the previous multi-measurement logic. This logic also unifies
the single-measurement and the multi-measurement implementations. `Result`
offers the same functionality, but it stores `Branch` objects instead of just a
`State` object and a list of samples.

Breaking changes:
 - A calculation function (which has to be registered in `_instruction_map`)
   now needs to return with a list of `Branch` objects, instead of a single
   `Result` object. This is only relevant for a user using the API directly.
 - During exection, the list of instructions were modified previously for
   resolving `modes`, when it is not specified. With this change, the `modes`
   parameter for each instruction is left untouched.
 - The list of samples got a fixed type `List[Tuple[Union[int, float], ...]]`
   (previously one could get numpy arrays).


piquasso/api/branch.py
```diff
@@ -0,0 +1,102 @@
...
+
+class Branch:
+    r"""Represents a branching in the computation caused by a measurement.
+
+    Describes a branch in the computation tree that arises due to a measurement
+    operation. Each branch contains the post-measurement quantum state, the outcomes
+    of the measurements that led to this branch, and the frequency (or weight) of
+    this branch in the overall ensemble. The frequency indicates how often this
+    particular outcome occurred relative to the total number of shots (simulations)
+    performed.
```
antalszava: Two things could elevate this explanation:
1. What an ensemble is in this context (I'm not sure all readers would be familiar with it)
2. The connection between frequency and probability.

piquasso/api/branch.py
```diff
@@ -0,0 +1,102 @@
...
+                style="width: 35%; max-width: 500px; height: auto;"
+                alt="Branch illustration"
+            >
+        </div>
+
+    In this case, two branches are created, one for each possible outcome of the
+    measurement. The branches contain the post-measurement quantum states
+    :math:`|\psi_0\rangle` and :math:`|\psi_1\rangle`, the outcomes 0 and 1, and
+    their corresponding probabilities :math:`p_0` and :math:`p_1`,
+    corresponding to each outcome. Since the ciruit is executed `shots` times, instead
```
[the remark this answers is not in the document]
antalszava: Imho, the following description could make it very explicit what's happening: "Each branch contains the post-measurement quantum state, the measurement outcome and the probability with which the specified outcome occurs. In our example, one branch contains :math:`|\psi_0\rangle`,  outcome 0 and probability :math:`p_0`, while the other contains :math:`p_1`, outcome 1 and probability :math:`p_1`."

piquasso/api/branch.py
```diff
@@ -0,0 +1,102 @@
...
+        has no associated quantum state (e.g., all the modes are measured), then
+        this is set to `None`.
+    :ivar outcome:
+        The accumulated outcome corresponding to the measurement(s) that caused the
+        branching. If no measurements were performed, then this is is set to `None`.
+    :ivar frequency:
+        The contribution of the state in the final ensemble. If the branch is the result
+        of a measurement with multiple possible outcomes, then this is set to the
+        ratio of the number of times `outcome` appears in the simulation, divided by
+        `shots` (which is the total number of samples taken). If the branch is not the
```
antalszava: I think we discussed during our conversations that the denominator of frequency is not always all the shots - has this changed, or am I misremembering something?

piquasso/api/result.py
```diff
@@ -13,36 +13,111 @@
...
 
-    def __init__(
-        self,
-        state: Optional[State] = None,
-        samples: Optional[Union[List[Tuple], "np.ndarray"]] = None,
-    ) -> None:
+class Result:
+    r"""Collects the results of a quantum computation.
+
+    The :class:`~piquasso.api.result.Result` class contains the results of a quantum
```
antalszava: Minor: why reference the class if this is already the docstring of the said class?

piquasso/api/simulator.py
```diff
@@ -239,56 +230,57 @@ def _delete_modes_from_active(active_modes, modes):
...
-    def _execute_multi_measurement_instructions(self, state, instructions, shots):
-        samples = []
+    @staticmethod
+    def _timestep(branches, calculation, instruction, shots):
+        new_branches = []
 
-        for _ in range(shots):
-            active_modes = tuple(range(self.d))
+        for branch in branches:
+            subbranches = calculation(
```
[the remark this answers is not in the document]
antalszava: Hmm, `implementation` also sounds quite general, did we somehow converge to `simulation_step` or so?

PR #562, opened 2026-03-16
Fix typo in heterodyne measurement docstring

mgarnier59:
This PR to quick fix a small typo found when reading the Piquasso docs. 

I will continue investigating the repo to see if we can make good use of it :).

Thanks!


Kolarovszki (approved):
Thank you so much! If you have any question, or if you want to see something implemented, feel free to reach out!

PR #570, opened 2026-04-10
TorchConnector

Tomev:
The implementation of basic TorchConnector. Passes all the tests that the `TensorflowConnector` passes.

Some of the tests and Takagi decomposition was adjusted, so that it uses more `connector.np` functions instead of e.g. in-place matrix multiplication. This allows us to avoid some of the `torch.Tensor` induced issues.

### DIFF
--- a/piquasso/_simulators/connectors/torch_/np_mock.py
+++ b/piquasso/_simulators/connectors/torch_/np_mock.py
@@ -69,3 +69,7 @@ class MockNumpy:
         dtype: torch.dtype = torch.promote_types(a.dtype, b.dtype)
 
         return a.to(dtype) @ b.to(dtype)
+
+    @staticmethod
+    def identity(n):
+        return torch.eye(n)
--- a/piquasso/_simulators/connectors/torch_/np_mock.py
+++ b/piquasso/_simulators/connectors/torch_/np_mock.py
@@ -15,46 +15,31 @@
 
 # This module contains mock-ups used by the TorchConnector.
 
+from typing import Any
+
 import torch
 
 
-def sum(tensor):
-    return torch.sum(tensor)
+class MockNumpy:
+    """A mock-up class implementing torch versions of the
+    functions used by piquasso that can be usually found
+    in numpy. This class will be used by TorchConnector."""
 
+    def __getattribute__(self, name: str, /) -> Any:
+        """
+        NOTE: Created to reduce boilerplate torch calls.
+        """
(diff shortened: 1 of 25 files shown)


tests/_math/test_decompositions.py
```diff
@@ -86,17 +93,20 @@ def test_takagi_on_real_symmetric_3_by_3_matrix(connector):
...
+    assert connector.np.allclose(connector.np.abs(singular_values), singular_values)
+    assert connector.np.allclose(
+        matrix, np.matmul(unitary, connector.np.diag(singular_values)) @ unitary.T
+    )
 
 
-@pytest.mark.parametrize(
-    "connector", [NumpyConnector(), JaxConnector(), lf("tensorflow_connector")]
-)
+# TODO(TR): This had no "lf" in ("tensorflow_connector"). Is that a problem?
```
Kolarovszki: I don't understand this.

piquasso/_simulators/connectors/torch_/np_mock.py
```diff
@@ -0,0 +1,91 @@
...
+
+# This module contains mock-ups used by the TorchConnector.
+
+from typing import Any
+
+import numpy as np
+import torch
+
+
+class MockNumpy:
```
Kolarovszki: It's not really a mock, is it? Wouldn't `Adapter` be a better term for this? For example, can we call this `TorchNumpyAdapter`?

Kolarovszki (changes requested):
Question: so if `TorchConnector` is not listed as an accepted connector for any simulator, then it is not needed to really implement all these methods like the `sqrtm`, right? Or do you plan to extend this to simulators as well?

But overall, I think this goes in a very good direction!

piquasso/_simulators/connectors/torch_/np_mock.py
Tomev: Adapter is better name. I'll change it in the next commit.

tests/_math/test_decompositions.py
Tomev: In the previous version of the file, all the parametrizations were specified directly with `@pytest.mark.parametrize`, not with `for_all_connectors`. Because of that some of the tests used `("tensorflow_connector")]` and the others had `lf("tensorflow_connector")`. Now all of the tests use `lf(...)` and I marked the tests that previously didn't with this note. 

If it doesn't matter I can drop the notes.

PR #574, opened 2026-05-05
feat(_simulators): `simulate` function

### LINKED ISSUE #573: `simulate` function
### What should be added?

We can help users by automatically choosing the simulator (e.g., `GaussianSimulator`, `FockSimulator`) through a function which could be conveniently called `simulate`, which would get a program, and optionally a connector and a config, and would return with the results.

Kolarovszki:
When users start to use Piquasso, they usually encounter the problem of choosing the correct simulator for their computations, which might be difficult to make.

To lighten this burden on the users, a lightweight `simulate` function is written, which is made available as `pq.simulate`. This function automatically chooses the simulator (`GaussianSimulator`, `SamplingSimulator`, `PureFockSimulator` or `FockSimulator`) and executes the instructions, returning with the simulation result.

Resolves #573

### DIFF
--- a/piquasso/_simulators/_simulate.py
+++ b/piquasso/_simulators/_simulate.py
@@ -1,0 +1,115 @@
+#
+# Copyright 2021-2026 Budapest Quantum Computing Group
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional, Union
+
+from .gaussian import GaussianSimulator
+from .fock import FockSimulator, PureFockSimulator
+from .sampling import SamplingSimulator
+
+from ..api.exceptions import NotImplementedCalculation
+from ..api.connector import BaseConnector
+from ..api.program import Program
+from ..api.config import Config
+from ..api.result import Result
+
+
+def _resolve_simulator_class(program, connector):
+    simulator_classes = (
+        GaussianSimulator,
+        SamplingSimulator,
+        PureFockSimulator,
+        FockSimulator,
+    )  # NOTE: Order matters, as the first compatible simulator is selected.
+
+    possible_simulator_classes = []
+
+    for simulator_class in simulator_classes:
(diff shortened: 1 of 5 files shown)


piquasso/_simulators/_simulate.py
```diff
@@ -0,0 +1,125 @@
...
+
+from ..api.exceptions import NotImplementedCalculation
+from ..api.connector import BaseConnector
+from ..api.program import Program
+from ..api.config import Config
+from ..api.result import Result
+
+
+def _resolve_simulator_class(program, connector):
+    simulator_classes = (
```
antalszava, line 30: It's worth mentioning here that the order is very strict and that it is going from the most specialized simulator for a specific task towards the most general simulator. This is important for future maintenance of this tuple.

piquasso/_simulators/_simulate.py
```diff
@@ -0,0 +1,125 @@
...
+
+    simulator_class = _resolve_simulator_class(program, connector)
+
+    simulator_kwargs = {
+        key: value
+        for key, value in {
+            "d": number_of_modes,
+            "connector": connector,
+            "config": config,
+        }.items()
```
antalszava: minor: These lines are a bit hard to read, imho the dict used here could be organized before `simulator_kwargs`

piquasso/_simulators/_simulate.py
```diff
@@ -0,0 +1,125 @@
...
+            instead of finite-shot sampling.
+
+    Returns:
+        Result: The result of the simulation.
+    """
+
+    simulator_class = _resolve_simulator_class(program, connector)
+
+    simulator_kwargs = {
+        key: value
```
antalszava: Is there no way to set the potential cutoff value?

tests/_simulators/test_simulate.py
```diff
@@ -0,0 +1,125 @@
...
+    assert type(result.state) is pq.GaussianState, (
+        "The state should be a GaussianState, as the program only contains "
+        "instructions compatible with Gaussian states."
+    )
+
+
+def test_simulate_passive_linear_gates():
+    with pq.Program() as program:
+        pq.Q() | pq.StateVector([1, 1, 0])
+
```
antalszava, line 41: Perhaps adding a PhaseShifter would strengthen this test as it doesn't only test BS gate but also PS

antalszava (commented):
Overall this is looking good, only main question I have is re. how cutoff can be set as kwarg to `simulate`. The rest are minor comments. Testing differentiability could be good too.

piquasso/_simulators/_simulate.py
Kolarovszki: I don't understand this question: `cutoff` can be set simply through `Config`.

antalszava: Ahh, good point!

antalszava (approved):
Looks good to me! 🎉

PR #586, opened 2026-06-03
Respect modes in SamplingSimulator particle measurements

### LINKED ISSUE #499: Always all modes are measured for ``ParticleNumberMeasurement`` on ``SamplingSimulator``
### Environment

- **Python version**: 3.12.7
- **Operating system**: WSL

### What is the error?

```py
import numpy as np
import piquasso as pq

with pq.Program() as program:

res.samples
```
```
```

### What should happen?

```
[(1, 1), (1, 1), (1, 1)]
```

DoView1:
## Summary
- trim `SamplingSimulator` particle-number samples to the requested measurement modes before building result branches
- add a regression test covering `pq.Q(0, 1)` and reversed `pq.Q(1, 0)` mode order

Fixes #499.

## Validation
- repro script for issue #499 now returns `[(1, 1), (1, 1), (1, 1)]`
- `tests/_simulators/sampling/test_measurements.py -q`: `35 passed, 1 warning`
- non-JAX sampling subset (`test_measurements.py`, `test_channels.py`, `test_preparations.py`, `test_state.py`): `60 passed, 1 warning`
- `python -m py_compile piquasso\_simulators\sampling\simulation_steps.py tests\_simulators\sampling\test_measurements.py`
- `black --check` on changed files
- `flake8` on changed files
- `git diff --check`

Note: collecting the broader sampling `test_gates.py` file in my local Windows environment requires the optional `jax` extra and failed before running tests because `jax` is not installed.

AI disclosure: I used OpenAI Codex to help trace the simulator path, implement the small patch, and run validation.

### DIFF
--- a/piquasso/_simulators/sampling/simulation_steps.py
+++ b/piquasso/_simulators/sampling/simulation_steps.py
@@ -249,6 +249,7 @@ def particle_number_measurement(
         rng=rng,
         postselect_data=postselect_data,
     )
+    samples = [tuple(sample[mode] for mode in instruction.modes) for sample in samples]
 
     binned_samples = get_counts(samples)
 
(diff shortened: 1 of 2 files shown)


Kolarovszki:
Sure, but this is a very naive solution and I think there should be a better algorithm for this case. Let me get back to you with some details soon.

PR #587, opened 2026-06-03
Add Gaussian ladder string moments

### LINKED ISSUE #524: `Gaussian.get_xp_string_moment` but for ladder operators
### What should be added?

Calculate the moments corresponding to a product of ladder (creation and annihilation) operators, similarly as described in [`GaussianState.get_xp_string_moment`](https://piquasso.readthedocs.io/en/stable/states/gaussian.html#piquasso._simulators.gaussian.state.GaussianState.get_xp_string_moment).

codexbountylab:
Closes #524.

## Summary

- add `GaussianState.get_ladder_string_moment`
- preserve the requested annihilation/creation index convention and operator order
- evaluate arbitrary-order Gaussian moments using ordered second moments and Isserlis' theorem
- add coverage for vacuum, coherent, multimode, CCR, `hbar` independence, and JAX/JIT

## Validation

- `python -m black --check piquasso/_simulators/gaussian/state.py tests/_simulators/gaussian/test_state.py`
- `python -m flake8 piquasso/_simulators/gaussian/state.py tests/_simulators/gaussian/test_state.py`
- `git diff --check`
- `pytest tests/_simulators/gaussian/test_state.py -k "get_xp_string_moment or get_ladder_string_moment" -q` (`18 passed`)
- explicit comparison against the XP expansion for every ladder-operator string through degree 6 on a nontrivial two-mode Gaussian state

## AI disclosure

I used OpenAI Codex to help inspect the codebase, brainstorm the ordered-moment transformation, and draft tests. I independently verified and refined the implementation using the checks listed above.

### DIFF
--- a/piquasso/_simulators/gaussian/state.py
+++ b/piquasso/_simulators/gaussian/state.py
@@ -1153,6 +1153,28 @@ class GaussianState(State):
         corresponds to
         :math:`\langle \hat{a}_0^\dagger \hat{a}_0 \rangle`.
 
+        The ordered first and second ladder-operator moments are obtained from their
+        quadrature counterparts using
+
+        .. math::
+            \boldsymbol{\xi} = \frac{1}{\sqrt{\hbar}} W \boldsymbol{Y},
+            \qquad
+            \left\langle \boldsymbol{\xi} \boldsymbol{\xi}^T \right\rangle
+            = \frac{1}{\hbar} W
+                \left\langle \boldsymbol{Y} \boldsymbol{Y}^T \right\rangle W^T,
+
+        where :math:`\boldsymbol{\xi} = (a_0, \dots, a_{d-1},
+        a_0^\dagger, \dots, a_{d-1}^\dagger)^T`,
+        :math:`\boldsymbol{Y}` is in xxpp-ordering, and
+
+        .. math::
+            W = \frac{1}{\sqrt{2}}
+                \begin{bmatrix} I & iI \\ I & -iI \end{bmatrix}.
+
+        Since the ordered quadrature second moments include the canonical commutation
+        relations, this transformation preserves the specified operator ordering.
+        Higher-order moments then follow from Isserlis' theorem.
+
         Args:
             string: The indices of the ladder operators.
 
--- a/piquasso/_simulators/gaussian/state.py
+++ b/piquasso/_simulators/gaussian/state.py
@@ -27,7 +27,11 @@ from piquasso._math.linalg import (
     is_symmetric,
     is_positive_semidefinite,
 )
-from piquasso._math.symplectic import symplectic_form, xp_symplectic_form
+from piquasso._math.symplectic import (
+    symplectic_form,
+    xp_symplectic_form,
+    complex_symplectic_form,
+)
 from piquasso._math.fock import get_fock_space_basis
 from piquasso._math.transformations import xxpp_to_xpxp_indices, xpxp_to_xxpp_indices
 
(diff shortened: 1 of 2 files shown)


codexbountylab:
AI disclosure: I used OpenAI Codex to help inspect the codebase, brainstorm the ordered-moment transformation, and draft tests. I independently verified the implementation against explicit XP expansions for every ladder-operator string through degree 6 on a nontrivial two-mode Gaussian state, and ran the project format/lint checks.

piquasso/_simulators/gaussian/state.py
```diff
@@ -1138,6 +1138,90 @@ def recursive(op_list: List[int]) -> complex:
...
+        transformation = np.block(
+            [[identity, 1j * identity], [identity, -1j * identity]]
+        ) / np.sqrt(2)
+
+        ordered_xp_covariance = (
+            self.xxpp_covariance_matrix / 2 + 0.5j * hbar * xp_symplectic_form(d)
+        )
+
+        first_order_moments = self.complex_displacement
+        second_order_moments = (
```
Kolarovszki, line 1204: Isn't the method `complex_covariance` useful for this?

Kolarovszki (commented):
Thank you! Overall, the PR looks good, I just have one minor question.

piquasso/_simulators/gaussian/state.py
codexbountylab: Good question. complex_covariance is related, but I don't think it can be used directly here because it is the symmetrized/Hermitian complex covariance, while this method needs ordered second moments of the ladder-operator vector itself.

For example, in the vacuum state we need to preserve the distinction between <a a†> = 1 and <a† a> = 0. The complex_covariance matrix does not encode that operator-ordering distinction by itself; the ordered quadrature second moments include the CCR term, and the W @ ordered_xp_covariance @ W.T / hbar transform preserves the requested order. That is why this implementation builds the ordered second-moment matrix explicitly instead of reading from complex_covariance.

AI disclosure: I used OpenAI Codex to help double-check this response against the implementation and the existing complex_covariance definition.

Kolarovszki: Yes, but isn't it `second_order_moments = (self.complex_covariance + 1j * complex_symplectic_form(d))/2`? I thought something like this would work.

codexbountylab: Yes, using complex_covariance works here. I pushed 7ca4cdf to use it.

The exact expression is slightly different from the sketch because complex_covariance uses xi_j^dagger in the second index, while this method needs ordered contractions with xi_j. So the implementation now uses

second_order_moments = (self.complex_covariance + complex_symplectic_form(d)) @ ladder_operator_swap / 2

where ladder_operator_swap converts the second index from the xi^dagger convention to the xi convention. That also means there is no extra 1j factor. I checked the expression against the previous xxpp-based formula for d = 1, 2, and 3 up to numerical precision.

Kolarovszki: Oh yes, I see! That's a good point, so it is not just a basis change, but almost. Otherwise, this function and the XP version is pretty much the same, right? To avoid repeating ourselves, is there a way to extract a common private method/function from these two implementations?

codexbountylab:
Thanks, that makes sense. I pushed 39e6e77 to extract the shared Isserlis recursion into a private `_get_operator_string_moment` helper.

Now `get_xp_string_moment` and `get_ladder_string_moment` each only construct their own ordered first- and second-order moments, then delegate the common pairing/singleton expansion to the helper.

Validation after the refactor:
- `python -m black --check piquasso/_simulators/gaussian/state.py`
- `python -m flake8 piquasso/_simulators/gaussian/state.py`
- `python -m py_compile piquasso/_simulators/gaussian/state.py`
- `git diff --check`

I could not rerun the pytest subset locally because the available Python is 3.13 while the existing local `.pqtest` dependency directory contains Python 3.12 NumPy extension modules, but this is a refactor of the previously tested recursion path and CI should exercise the same tests.

Kolarovszki (approved):

codexbountylab:
Hi @Kolarovszki, I noticed #587 was closed because it appeared that I was not registered for unitaryHACK.

I wanted to clarify that I did receive a unitaryHACK26 registration confirmation email, and my registration was submitted with the GitHub account `codexbountylab`.

I have contacted the unitaryHACK team to confirm whether there is any mismatch on their side. I’ll follow up once I hear back.

Kolarovszki:
Hi @codexbountylab, I am deeply sorry, I closed your PR by mistake (I misread some info about banned accounts sent to the maintainers).

codexbountylab:
> Hi @codexbountylab, I am deeply sorry, I closed your PR by mistake (I misread some info about banned accounts sent to the maintainers).

No worries, thanks for clarifying.

PR #590, opened 2026-06-04
Measure only the specified modes for ParticleNumberMeasurement on SamplingSimulator

### LINKED ISSUE #499: Always all modes are measured for ``ParticleNumberMeasurement`` on ``SamplingSimulator``

sarah-ceo:
Closes #499 

## What
I've added the ability to measure only the specified modes for ParticleNumberMeasurement on SamplingSimulator, as was requested in the issue.

## How
I used the same logic as in the PureFockSimulator since that one was working correctly: I added a method to the SamplingState that creates a reduced version of the state with only the selected modes, and this method is called in the SamplingSimulator's particle_number_measurement function.

## Testing
I added a function in the SamplingSimulator's test_measurements that tries the example given in the related issue.

## Notes
I used Claude's assistance to understand how the properties of the SamplingState should be modified, especially the interferometer.

### DIFF
--- a/piquasso/_simulators/sampling/utils.py
+++ b/piquasso/_simulators/sampling/utils.py
@@ -183,7 +183,7 @@ def generate_samples(
     k = len(selected_modes)
 
     if k == 1 and n>=3 and n<10:
-        return _generate_marginal_samples(
+        return _generate_k_modes_samples(
             input, 
             shots, 
             interferometer, 
@@ -487,7 +487,7 @@ def _calculate_singular_values_matrix_expansion(singular_values_vector):
     return np.diag(expansion_values)
 
 
-def _generate_marginal_samples(
+def _generate_k_modes_samples(
         input, shots, interferometer, rng, selected_modes, n
 ):
     '''
@@ -496,12 +496,12 @@ def _generate_marginal_samples(
     samples = []
 
     while len(samples) < shots:
-        sample = _generate_marginal_sample(selected_modes, n, input, interferometer, rng)
+        sample = _generate_k_modes_sample(selected_modes, n, input, interferometer, rng)
         samples.append(tuple(sample))
 
     return samples
 
-def multiply_polynomials(poly_arr_shape, arr):
+def _multiply_polynomials(poly_arr_shape, arr):
     '''
         Multiplies arrays of polynomial coefficients with fft.
     '''
@@ -510,7 +510,7 @@ def multiply_polynomials(poly_arr_shape, arr):
     ifft_arr = np.fft.ifftn(prod_arr, axes=range(1, len(prod_arr.shape)))
     return ifft_arr
 
-def compute_laguerre(k, x, ra):
+def _compute_laguerre(k, x, ra):
     '''
         Computes the Laguerre series at degree ra and returns an array of its polynomials.
     '''
@@ -535,7 +535,7 @@ def compute_laguerre(k, x, ra):
         coeffs[:, i, relevant_indices] = poly_coeffs
     return coeffs
 
-def _generate_marginal_sample(modes, n, inputs, interferometer, rng):
+def _generate_k_modes_sample(modes, n, inputs, interferometer, rng):
     '''
         Implements sampling when k-modes have been selected for measurement.
     '''
@@ -556,8 +556,8 @@ def _generate_marginal_sample(modes, n, inputs, interferometer, rng):
     faw = np.conj(V).T[:, None, :]*np.power(omegas, -1)[None,...]
     xa = np.einsum('nmi,nmj->mnij', faz, faw)
     xa = np.sum(-xa, axis=2)
-    l_ra = compute_laguerre(k, xa, occupied_inputs)
-    P_omega = multiply_polynomials(poly_arr_shape, l_ra)
+    l_ra = _compute_laguerre(k, xa, occupied_inputs)
+    P_omega = _multiply_polynomials(poly_arr_shape, l_ra)
     D_t = np.sum(P_omega, axis=0)
     D_t /= N**k
 
--- a/piquasso/_simulators/sampling/utils.py
+++ b/piquasso/_simulators/sampling/utils.py
@@ -25,6 +25,7 @@ from piquasso._math.combinatorics import partitions, partitions_bounded_k
 from piquasso._math.fock import get_fock_space_basis
 from .marginal import (
     _get_diagonal_coefficients,
+    cutoff_fock_space_dim_array,
     _probabilities_from_diagonal_coefficients,
 )
 
@@ -601,10 +602,14 @@ def _generate_marginal_samples(
                     ]
                     intermediate_diagonal = diagonal_coefficients[relevant_indices]
 
+                indices = cutoff_fock_space_dim_array(
+                    np.arange(n + 2), num_postselected_modes + 1
+                )
                 probabilities = _probabilities_from_diagonal_coefficients(
                     intermediate_diagonal,
                     intermediate_compositions,
                     outcomes_with_postselection,
+                    indices,
                 )
                 if np.all(probabilities == 0):
                     break
--- a/piquasso/_simulators/sampling/utils.py
+++ b/piquasso/_simulators/sampling/utils.py
@@ -22,13 +22,6 @@ from piquasso.api.exceptions import InvalidSimulation
 
 from piquasso._math.combinatorics import partitions, partitions_bounded_k
 
-from piquasso._math.fock import get_fock_space_basis
-from .marginal import (
-    _get_diagonal_coefficients,
-    cutoff_fock_space_dim_array,
-    _probabilities_from_diagonal_coefficients,
-)
-
 """
 This is a contribution from `theboss`, see https://github.com/Tomev-CTP/theboss.
 
@@ -158,8 +151,6 @@ def generate_samples(
     rng,
     reject_condition,
     postselect_data,
-    selected_modes,
-    lossy=False,
 ):
     """
     Generates samples corresponding to the Clifford & Clifford algorithm B from
@@ -185,11 +176,8 @@ def generate_samples(
     Returns:
         The generated samples.
     """
-    if 0 < len(selected_modes) < len(input) and not lossy:
-        return _generate_marginal_samples(
-            input, interferometer, selected_modes, shots, rng, postselect_data
-        )
-    elif len(postselect_data[0]) > 0:
+
+    if len(postselect_data[0]) > 0:
         sample_generator = partial(
             _generate_sample_with_postselect,
             reject_condition=reject_condition,
@@ -205,7 +193,6 @@ def generate_samples(
         sample_generator=sample_generator,
         interferometer=interferometer,
         rng=rng,
-        selected_modes=selected_modes,
     )
 
 
@@ -216,7 +203,6 @@ def generate_lossy_samples(
     interferometer_svd,
     rng,
     postselect_data,
-    selected_modes,
 ):
     """
     Basically the same algorithm as in `generate_samples`, but doubles the system size
@@ -239,8 +225,6 @@ def generate_lossy_samples(
         rng,
         reject_condition=lambda: False,
         postselect_data=postselect_data,
-        selected_modes=selected_modes,
-        lossy=True,
     )
 
     # Trim output state
@@ -262,13 +246,7 @@ def _get_first_quantized(occupation_numbers):
 
 
 def _generate_samples(
-    input,
-    shots,
-    calculate_permanent_laplace,
-    interferometer,
-    sample_generator,
-    rng,
-    selected_modes,
+    input, shots, calculate_permanent_laplace, interferometer, sample_generator, rng
 ):
     d = len(input)
     n = np.sum(input)
@@ -287,7 +265,7 @@ def _generate_samples(
             rng=rng,
         )
         samples.append(tuple(sample))
-    samples = [tuple(sample[mode] for mode in selected_modes) for sample in samples]
+
     return samples
 
 
@@ -491,143 +469,3 @@ def _calculate_singular_values_matrix_expansion(singular_values_vector):
     expansion_values = np.sqrt(vector_of_squared_expansions)
 
     return np.diag(expansion_values)
-
-
-def _generate_marginal_samples(
-    initial_state, interferometer, selected_modes, shots, rng, postselect_data
-):
-    """
-    Implements mode-by-mode sampling with postselection.
-    """
-    n = sum(initial_state)
-    k = len(selected_modes)
-
-    if n == 0:
-        return [(0,) * k] * shots
-
-    postselect_modes, postselect_photons, max_sample_generation_trials = postselect_data
-    selected_modes = np.array(selected_modes)
-    for postselected in postselect_modes:
-        selected_modes[selected_modes >= postselected] += 1
-    all_selected_modes = np.array(sorted(tuple(selected_modes) + postselect_modes))
-
-    compositions = get_fock_space_basis(len(all_selected_modes), n + 1)
-    diagonal_coefficients = _get_diagonal_coefficients(
-        input_photons=initial_state,
-        interferometer=interferometer,
-        all_modes=all_selected_modes,
-        compositions=compositions,
-    )
-
-    samples = []
-    while len(samples) < shots:
-        sample = []
-
-        postselections = {
-            postselect_modes[i]: postselect_photons[i]
-            for i in range(len(postselect_data[0]))
-        }
-
-        retry = True
-        trial_count = 0
-
-        while retry:
-            trial_count += 1
-
-            if trial_count > max_sample_generation_trials:
-                raise InvalidSimulation(
-                    "Too many trials during sample generation. "
-                    "The specified postselection criteria may be very highly unlikely. "
-                    "Aborting. "
-                    "To increase the limit, set Config.max_sample_generation_trials "
-                    f"to a higher value. Current value: {max_sample_generation_trials}."
-                )
-
-            for mode in selected_modes:
-
-                if mode in postselections:
-                    continue
-
-                cutoff = int(n - np.sum(list(postselections.values())) + 1)
-                outcomes = get_fock_space_basis(1, cutoff)
-                num_postselected_modes = len(postselections.keys())
-                ordered_modes = sorted(list(postselections.keys()) + [mode])
-                current_index = ordered_modes.index(mode)
-
-                outcomes_with_postselection = np.zeros(
-                    (len(outcomes), num_postselected_modes + 1), dtype=np.int64
-                )
-                if num_postselected_modes:
-                    postselected_array = np.array(
-                        [list(postselections.keys()), list(postselections.values())],
-                        dtype=int,
-                    )
-                    postselected_array[0, postselected_array[0, :] < mode] = range(
-                        current_index
-                    )
-                    postselected_array[0, postselected_array[0, :] > mode] = range(
-                        current_index + 1, num_postselected_modes + 1
-                    )
-                    outcomes_with_postselection[:, postselected_array[0]] = np.asarray(
-                        postselected_array[1], dtype=np.int64
-                    )
-                outcomes_with_postselection[:, current_index : current_index + 1] = (
-                    outcomes
-                )
-
-                intermediate_compositions = get_fock_space_basis(
-                    num_postselected_modes + 1, n + 1
-                )
-                intermediate_diagonal = diagonal_coefficients
-                if intermediate_compositions.shape[1] < len(all_selected_modes):
-                    intermediate_compositions_full = np.zeros(
-                        (intermediate_compositions.shape[0], len(all_selected_modes)),
-                        dtype=int,
-                    )
-                    columns_ordering = [
-                        i
-                        for i in range(len(all_selected_modes))
-                        if all_selected_modes[i] in ordered_modes
-                    ]
-                    intermediate_compositions_full[:, columns_ordering] = (
-                        intermediate_compositions
-                    )
-                    intermediate_compositions_dict = {
-                        tuple(row) for row in intermediate_compositions_full
-                    }
-                    relevant_indices = [
-                        i
-                        for i, row in enumerate(compositions)
-                        if tuple(row) in intermediate_compositions_dict
-                    ]
-                    intermediate_diagonal = diagonal_coefficients[relevant_indices]
-
-                indices = cutoff_fock_space_dim_array(
-                    np.arange(n + 2), num_postselected_modes + 1
-                )
-                probabilities = _probabilities_from_diagonal_coefficients(
-                    intermediate_diagonal,
-                    intermediate_compositions,
-                    outcomes_with_postselection,
-                    indices,
-                )
-                if np.all(probabilities == 0):
-                    break
-                probabilities /= np.sum(probabilities)
-
-                n_photons_list = outcomes.flatten()
-                n_photons_probs = probabilities.real
-                positive_probs = np.where(n_photons_probs > 0)
-
-                n_photons = rng.choice(
-                    n_photons_list[positive_probs], p=n_photons_probs[positive_probs]
-                )
-
-                sample.append(n_photons)
-                postselections[mode] = n_photons
-
-            if len(sample) > 0:
-                retry = False
-
-        samples.append(tuple(sample))
-    return samples
--- a/piquasso/_simulators/sampling/utils.py
+++ b/piquasso/_simulators/sampling/utils.py
@@ -13,12 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import defaultdict
 from itertools import product
 import math
 import numpy as np
-from scipy.special import factorial
-from numpy.polynomial.laguerre import lag2poly
+from scipy.special import factorial, comb
 
 from functools import partial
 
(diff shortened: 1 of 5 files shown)


piquasso/_simulators/sampling/state.py
```diff
@@ -431,3 +431,23 @@ def __eq__(self, other: object) -> bool:
             and self.is_lossy == other.is_lossy
             and self._postselections == other._postselections
         )
+    
+    def reduced(self, modes: Tuple[int, ...]) -> "SamplingState":
```
Kolarovszki: I am very doubtful that this method does what it promises. Can you explain your line of thought?

Unfortunately, it is not so simple that one can just crop the interferometer and the input state and call it a day. This can be seen very easily, by considering the case of a $\ket{11}$ input state scattered by a 50/50 beamsplitter.

sarah-ceo: Hello, sure: from what I understood, the ParticleNumberMeasurement needed to sample from a state that should consider only a subset of the modes. So, where the FockSimulator would reduce the density matrix, I thought we needed to consider only a sub-part of the interferometer. Now that I’ve looked more into it, I realize that reducing the state before sampling is probably not a good approach at all with a SamplingState that doesn’t store anything like a full density matrix… Should we just keep the relevant modes after sampling then?

And sorry I gave such an over-simplistic and naive solution: I am genuinely trying to learn and do not mean to waste your time at all. I will study this issue more to offer a better contribution.

Kolarovszki: Keeping the relevant modes after sampling would be a solution, but there is a better algorithm than that. I will edit the issue description accordingly and notify you, but I am a bit overwhelmed ATM.

sarah-ceo:
Hello @Kolarovszki,

I have implemented your algorithm, but expanding all the Laguerre polynomial coefficients takes a really long time as k grows...

It might just be the way I implemented it though, so I wanted to check with you. If I expand after each Laguerre and dynamically integrate the coefficients, do I still have to compute all the sub-combinations of alpha and beta?

For example, to update the coefficient P_{(2,1),(2,1)}, I currently do a sum over all the decompositions such as:
((1,0),(1,0)) + ((1,1),(1,1)),

((1,0),(0,1)) + ((1,1),(2,0)),

((0,1),(1,0)) + ((2,0),(1,1)),

((0,1),(0,1)) + ((2,0),(2,0)),

((2,1),(2,1)) + ((0,0),(0,0)),
etc., which is the most time-consuming part.

Am I really supposed to do this or did I misunderstand? 

If not, maybe I can look into the ways to either avoid expanding all the Laguerre polynomials, or look for a shortcut to compute only the diagonal coefficients, if there is one...

Thanks !

Kolarovszki:
@sarah-ceo 

Yes, unfortunately, it will blow up rapidly in $k$, and the algorithm is only efficient for small fixed $k$s. In this case, it is polynomial-time (although the polynomial degree grows with $k$), whereas the naive solution would be still exponential.

What you mention is the most time-consuming part of the algorithm. The question is how to compute $P_{\alpha, \alpha}$ as efficiently as possible? I suggest the following, although this might not be the best option:

We start from
```math
P(\mathbf{z}, \mathbf{w}) = \sum_{\alpha, \beta} P_{\alpha, \beta} $\mathbf{z}^{\alpha} \mathbf{w}^{\beta}$.
```
First, I think one can try to write $\mathbf{z}^{\alpha} \mathbf{w}^{\beta}$ as $\mathbf{t}^{\alpha} \mathbf{u}^{\beta-\alpha}$. Now we can keep the constant term in $\mathbf{u}$ in the expansion, which corresponds to the diagonal elements, and that only depends on $t$. Let us denote that by $D(t)$, and we know that
```math
D(\mathbf{t}) = \sum_{\alpha} P_{\alpha, \alpha} \mathbf{t}^\alpha.
```
If we could get this polynomial, we would be good. We know that $P$ has degree up to $n$ in the elements of $\mathbf{u}$ or $$\mathbf{u}^{-1}$, so let us take $N = n+1$. By averaging over the roots of unity, we get
```math
D(\mathbf{t}) = \frac{1}{N^k} \sum_{\omega_1^N = \dots = \omega_k^N = 1} P(\mathbf{t} \omega, \omega^{-1}).
```
(This is indeed excessive, but I haven't found any better solution for extracting $D(\mathbf{t})$ so far.)
Let us denote $P_\omega(\mathbf{t}) = \frac{1}{N^k} \sum_{\omega_1^N = \dots = \omega_k^N = 1} P(\mathbf{t} \omega, \omega^{-1})$, and we can symbolically compute $P_\omega$. We know that it has $\binom{n+k}{k}$, coefficients and multiplying the polynomials in $P_\omega(\mathbf{t})$ costs $nk \binom{n+k}{k}$.  Finally, we have to sum (n+1)^k of these polynomials. The coefficients of $D(\mathbf{t})$ will be the required $P_{\alpha, \alpha}$ constants in the probability. This algorithm has runtime $O(nk (n+1)^k \binom{n+k}{k})$, which seems bad, but not that bad for small $k$, and for fixed $k$ it is polynomial time in $n$ (whereas the naive algorithm would still be exponential in $n$).

Maybe there is a better solution out there, I don't know.

sarah-ceo:
Alright, thank you ! I'll get right on it.

sarah-ceo:
Hi @Kolarovszki,

I have tried implementing the solution but the expansion is still too costly, even with small k and n. I've seen the new PR which seems to completely bypass this and has performances way beyond my solution's reach.

So I think we can close my PR unfortunately. Thank you for your time and for the opportunity.

Best

Kolarovszki:
@sarah-ceo Oh! But that is very similar to what I have described here I think.

sarah-ceo:
@Kolarovszki 

Yes I think the implementation logic is correct: it has the same distribution as the post-sampling discarding method. But the expansion makes it much slower than computing all modes, even for small k and n.

I thought at first it might be the reliance on Sympy but even without it, it is still very slow: right now it takes 6.4 seconds for 1000 shots with k=2 and n=4, while the original solution takes 0.16 seconds. 

Even when I comment out almost everything in my solution and compute only the expansion for one possible_y, the code still takes 0.2 seconds: so computing the expansion once is slower than the whole original code. That's why I doubt I can reach good performances.

(I didn't integrate my solution fully but tested it locally simply by adding "sample = generate_k_modes_marginal_sample(range(2), n, input, interferometer, rng)" at utils.py:271 and commenting out 263:270.)

Kolarovszki:
@sarah-ceo I see, but I don't understand why this solution is so slow. Perhaps an FFT-based polynomial multiplication would speed it up (?)

sarah-ceo:
@Kolarovszki 

It is slow because I used dictionaries to multiply the polynomials: I had no recollection of fft-based polynomial multiplication. So yes, when I use that and compute Laguerre by recurrence, the speedup is huge thanks to vectorization.

It is still not faster than the original code at that point because the next bottleneck is my loop over possible ys. To get rid of that, I'd need to use a huge array to keep track of all the polynomials (now possible though thanks to the vectorization). 

And that's essentially what the other PR does.

Your solution is fast, it's just my implementation that is slow because it isn't vectorized.

Kolarovszki:
@sarah-ceo I agree, the other PR is a bit closer to the truth, but could also be improved. I think we could reach (or could have reached) to a good solution from this PR by iterating, but it might require lots of steps.

sarah-ceo:
@Kolarovszki 

I just don't want to do the same thing as the other PR, it would be a waste of time. If you think we can reach something different but as efficient with this solution, then I am ok to work more on it and iterate. Otherwise, I would rather tackle some of the other issues in Piquasso that are outside of the UH26.

(My goal here was just to learn more about photonic QC and how to contribute to open-source libraries, so I am happy to work on the less popular issues without it being a competition.)

Kolarovszki:
@sarah-ceo 

Sure! There are many issues in the issues page: https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues

For example this is also connected to the `SamplingSimulator`: https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues/600

Kolarovszki:
@sarah-ceo I closed the other PR due to the suspected violation of the AI policy: https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/603

So if you would like to work on this issue, feel free to reopen it.

sarah-ceo:
@Kolarovszki 

Oh alright then, sure ! I'll get back on it. Can you reopen it please? (I don't have to rights to, it seems)

sarah-ceo:
@Kolarovszki 

Thanks ! I've managed to vectorize quite a bit and bring down the computation time to 0.4 seconds for 1000 shots with k=2 and n=4. I am not sure whether I can gain much more speed, what do you think? 

Maybe I could remove the loop on a but I am not certain: I will look into it more tomorrow morning. 

I also still have to find the optimal k/n to handle the branching between the original solution and this one. Then add the post-sampling discarding method to the original solution and a pytest.

Do you have some corrections / recommendations on the functions so far?


Note: And the sample distributions for the two solutions look quite similar with k = 2 or k = 3, shots=1000 and different StateVectors.
The test code was:
with pq.Program() as program:
        pq.Q() | pq.StateVector(state_vector)
        pq.Q(0, 1) | pq.Beamsplitter()
        pq.Q(0, 2) | pq.Beamsplitter()
        pq.Q(2, 3) | pq.Beamsplitter()
        pq.Q(*selected_modes) | pq.ParticleNumberMeasurement()

k2-n4

k3-n8

sarah-ceo:
@Kolarovszki 

Hello,

So I did manage to remove the big loop on a and now the performances are interesting but only when k=1 and n>=3:

Figure_1

I integrated the solution for these numbers and I added the post-sampling discarding method to the original solution + a pytest.

There is still a problem however: when n>= 10, the probabilities do no longer add up to 1 in my solution. In fact, they completely blow up after n=15:
    # n = 1: sum = 1.0
    # n = 4: sum = 1.0000000000000044
    # n = 6: sum = 0.9999999999999999
    # n = 7: sum = 1.0000000000004894
    # n = 8: sum = 1.0000000000401201
    # n = 9: sum = 1.0000000000725184
    # n = 10: sum = 1.0000001111186574
    **# n = 14: sum = 1.0510241882921634
    # n = 15: sum = 7.744103882742677
    # n = 16: sum = 57.77057914562677
    # n = 18: sum = 2239.3859746861**

I guess it must be due to some numerical instability so I'll have to investigate it tomorrow... In the meantime I set the condition for the solution to be "if k == 1 and n>=3 and n<10" .

And sorry it is taking me so long !

Kolarovszki (changes requested):
Hi! Sorry for the late response.

The plots look promising indeed. Since my initial comment on the algorithm, I have iterated on the solution quite a bit.

Let me recap the theory. Let $T=\{\ell_1,\ldots,\ell_k\}$ be the subset of output modes we want to observe, and let the occupied input modes be $a=1,\ldots,s$, with occupations $r_a$ and $\sum_a r_a=n$. Define
```math
    V_{aj}=U_{\ell_j,a}^{*},
    \qquad
    f_a(\mathbf{z})=\sum_{j=1}^k V_{aj}z_j, \qquad  P(\mathbf{z},\mathbf{w})
    =
    \prod_{a=1}^s
    L_{r_a}\!\left(-f_a(\mathbf{z})\overline{f_a(\mathbf{w})}\right),
```
where $L_r$ is a Laguerre polynomial of degree $r$. Let us denote
```math
F_a(\mathbf{z}, \mathbf{w}) := L_{r_a}\!\left(-f_a(\mathbf{z})\overline{f_a(\mathbf{w})}\right)
```
for brevity. Now, expand $P$ as
```math
    P(\mathbf{z},\mathbf{w})
    =
    \prod_{a=1}^s F_a(\mathbf{z}, \mathbf{w}) 
    =
    \sum_{\alpha,\beta\in\mathbb N^k}
    P_{\alpha,\beta} \mathbf{z}^\alpha \mathbf{w}^\beta .
```
The claim is that the marginal probability of observing $\mathbf y$ in the selected modes is
```math
   \mathrm{Pr}(Y=\mathbf y)
    =
    \sum_{\substack{\alpha\ge \mathbf y \\ |\alpha| \leq n}}
    \alpha!\,
    P_{\alpha,\alpha}
    \prod_{j=1}^k
    \binom{\alpha_j}{y_j}
    (-1)^{\alpha_j-y_j}.
```

The problem reduces to computing $P_{\alpha,\alpha}$ for all $\alpha \in \mathbb{N}\_0^n$ such that $|\alpha| \leq n$. From these, we can straightforwardly compute the probabilities, and do a mode-by-mode boson sampling. So let us turn to computing $P_{\alpha,\alpha}$, which can be done by expanding $P$ up to to total degree $k$ in $\mathbf{z}$ and $\mathbf{w}$ variables. For this, we need to multiply the $F_a$ polynomials together.

An important observation that the part of $P$ that we need only contains $\binom{n+k}{k}^2$ coefficients (because we only need $P_{\alpha, \alpha}$ for $|\alpha| \leq k$ in the end), instead of $(n+1)^{2k}$. So, in theory, you can compute the $F_a$ polynomials one by one, and consecutively multiply them, but after each multiplication, you should be able to truncate to at most $\binom{n+k}{k}^2$ coeffs (of course, the intermediate polynomials need less memory depending on the total photon number). At the end, you can extract the diagonals.

There is some extra structure, so that we can reduce $\binom{n+k}{k}^2$ even further. We know that $F_a$ has total degree $r_a$ in each $\mathbf{z}$ and $\mathbf{w}$ and hence
```math
F_a(\mathbf{z}, \mathbf{w}) = \sum_{\substack{\alpha, \beta \geq 0\\ |\alpha| = |\beta| \leq r_a}} A_{\alpha, \beta} \mathbf{z}^\alpha \mathbf{w}^\beta,
```
i.e., there is also the $|\alpha| = |\beta|$ restriction. Importantly, this extends to their product as well, so it extends to $P$. This means that there are no "cross-coefficients", i.e., $P_{\alpha, \beta} = 0$ whenever $|\alpha| \neq |\beta|$, and for every intermediate product the same holds. Hence, $P$, $F_a$ and every intermediate decomposes into "bihomogeneous" polynomials. This means that a lot more coefficients are $0$, and whenever $k$ is small, it makes a significant difference if we avoid

How to do the multiplication? I don't think we need to store the polynomial in an array of size N^k, it is wasteful when $k > 1$, as there are many-many zeros inside the array. It is the easiest to do FFT if you represent the polynomials as rectangular arrays, but maybe not the most efficient. One can try using sparse representations and using sparse FFT, but there **should** be a nicer way I think. The standard tools from NumPy, SciPy won't cut it, one has to implement a custom, dynamic code for this: some truncation-aware polynomial multiplication based on FFT. One has to figure out how to multiply bihomogeneous polynomials as efficiently as possible. I'm reading this paper: https://arxiv.org/pdf/0901.4323 (an LLM suggested this of course), perhaps Section 4 is interesting for us.

In theory, this approach is good because it has runtime
```math
O(s \binom{n+k}{k}^2 \log \binom{n+k}{k}),
```
because the computation is dominated by the last multiplication, which has $n$ photons. This is improved by the homogeneity assumption, which yields a factor of $n$ speedup for fixed $k$ I think. The downside of this approach is that it uses $O(\binom{n+k}{k}^2)$ memory but I was thinking about it and it is not that bad in most cases, as $k$ is typically small.

The sum-over-grid strategy that you have implemented is also good, because it uses significantly less memory, but the runtime is worse, because it contains a $(n+1)^k$ factor due to the grid. So I would go with the strategy I just described.

I'm sorry, this turned out to be quite a difficult issue.

sarah-ceo:
Hello @Kolarovszki,

Sorry for the late answer: I've been working on the sparse polynomial multiplication algorithm. I've finally come to something that works but I still need to improve the speed to make it faster than the original solution (some loops need to be removed like last time).

I might be able to do it tomorrow, but there is still one big issue to work on: the numerical stability. It is worse than before when k>1, the probabilities stop adding up very quickly:
k = 2, n = 1, sum = 1.0
k = 2, n = 2, sum = 1.0
k = 2, n = 3, sum = 0.9999999999999991
k = 2, n = 4, sum = 5.250000000045578

It is more stable for k = 1 but I think I need to look into this... Let me know if you can think of where this might originate from, or if there are some parts of the algorithm that don't seem quite right to you so far.

Thanks !

Kolarovszki:
Hi @sarah-ceo!

Isn't this some kind of other error, like a normalization error? The almost exact value of 5.25 for k=2, n=4 doesn't seem like a numerical stability issue for me, considering that k=2, n=3 is very close to 1.

But I am very happy that you have implemented this multiplication algorithm by Joris van der Hoeven. It can happen that it is numerically unstable, but in our case, the coefficients come from unitary matrices, and I think the coefficients of $F_a$ should be bounded by $1$ when $a = 0$ or $1$, which is typically the case in Boson Sampling, so maybe the numerical instability is not that bad.

Kolarovszki:
@sarah-ceo Also, I will assign the issue to you, because you have come closest to the solution by implementing the sparse multiplication algorithm, and you have worked the most on this issue IMHO.

piquasso/_simulators/sampling/utils.py
```diff
@@ -469,3 +485,209 @@ def _calculate_singular_values_matrix_expansion(singular_values_vector):
...
+
+def _multiply_polynomials(all_ys, alphas_to_indices, indices_to_alphas, P, Q, N, k, d, return_dense=False):
+    '''
+        Multiplies arrays of polynomial coefficients using "Sparse polynomial multiplication" algorithm by Joris van der Hoeven.
+    '''
+    X, SP, SQ = _get_support_points(alphas_to_indices, N, k, d, P,Q)
+    points = _set_evaluation_points(all_ys, X, N, d)
+
+    V = _build_V_matrix(points)
+    V_inv = np.linalg.pinv(V)
```
Kolarovszki: So this inversion is problematic, as it makes the algorithm inefficient. Let me think how we can solve this.

sarah-ceo:
> Hi @sarah-ceo!
> 
> Isn't this some kind of other error, like a normalization error? The almost exact value of 5.25 for k=2, n=4 doesn't seem like a numerical stability issue for me, considering that k=2, n=3 is very close to 1.
> 
> But I am very happy that you have implemented this multiplication algorithm by Joris van der Hoeven. It can happen that it is numerically unstable, but in our case, the coefficients come from unitary matrices, and I think the coefficients of F a should be bounded by 1 when a = 0 or 1 , which is typically the case in Boson Sampling, so maybe the numerical instability is not that bad.

Oh I think you are right about the sum problem not being from numerical instability: I thought it was linked to the Vandermode matrix being ill-conditioned with very high T and too clustered points. But I just changed the max Kronecker exponent to go less high and the condition of the matrix reduced greatly (from 13816 to 304 for k = 2, n = 4) but the sum was still the same at 5.25... so I must have forgotten a normalization somewhere.

Looking into it.

sarah-ceo:
@Kolarovszki 

So actually the bug was from a mistake in the dense matrix rebuilding. I completely removed it since it served no purpose besides helping me keep track. Everything is just stored as dictionaries again.

I sped up the probabilities calculations and made polynomial multiplication work for a list instead of just 2, so now we are back into better computing time for k=1 compared to the original solution (and the distributions are still similar):

Figure_1

_______________

There is still **big** numerical instability though, and its threshold depends on the length of the state vector, its configuration and the modes measured...

State vector: [4, 4, 3]
k = 1, n = 11, sum = 1.0000000000000782

**State vector: [4, 4, 4]
k = 1, n = 12, sum = 1.0000000497460455**

...

**State vector: [5, 4, 4, 4, 4]
k = 1, n = 21, sum = 1.304115379429902**

**State vector: [5, 5, 4, 4, 4]
k = 1, n = 22, sum = 10.875151161332639**

...

**State vector: [1, 1, 0, 0, 0]
k = 3, n = 2, sum = 1.00153857577061**

**State vector: [1, 1, 1, 0, 0]
k = 3, n = 3, sum = 227935.39169003477**

It's very annoying because it forces me to reduce the scope of application of this solution to just k=1, n>3 and n<10...

Kolarovszki:
@sarah-ceo 

Maybe we should avoid this approach for now then. This might be very difficult to implement as an UH issue, and it is unclear if it's numerically stable or not. Also, van der Hoeven doesn't apply to multiplying bihomogeneous polynomials as in our case, and it is tricky to generalize for our case.

So this van der Hoeven algorithm works well over finite fields, and for integers if you choose large enough $p$ I suppose. For floating points, one can map the coefficients into integers, but this can introduce precision issues.

Also, applying the evaluation map should not be done by constructing the Vandermonde matrix explicitly, as there are better algorithms for computing the evaluation map and its inverse.

Would it be OK for you if I would push a (suboptimal) algorithm for computing the $k$-marginal probabilities, and leave optimization for future work?

sarah-ceo:
@Kolarovszki 

Ah, I see... too bad ! It was really making sense on paper. 

Sure, is there anything else I should/could do regarding this for now then? Should I revert to a suboptimal solution or create a new branch or something?

Kolarovszki:
@sarah-ceo 

Yeah, I think it is mentioned in the arxiv preprint:
image

Kolarovszki:
@sarah-ceo 
 
> Sure, is there anything else I should/could do regarding this for now then? Should I revert to a suboptimal solution or create a new branch or something?

I think if you finalize the PR with a naive algorithm for computing the k-marginal probabilities, that is fine as well, and later I will replace the implementation, if that's good (or, you can also replace it later).

sarah-ceo:
@Kolarovszki 

Alright, I removed my code and left only the post-sampling discarding and the test: is that alright?

I looked at the paper for the computation of E and E<sup>-1</sup> (Fast Modular Transforms via Division): it does seem rather complicated at first glance, certainly not something I could complete in the next few days. (Sorry I missed it earlier, I focused on the Vandermonde and inverse and did not quite catch that it wasn't the recommended solution.)

I don't mind if you want to replace the implementation later: I will focus on some other issues for the moment. But if you haven't had the chance to implement it in like two weeks, I can certainly come back to it ! Also, I wouldn't mind investigating the numerical stability, as it really bugs me.

Kolarovszki:
@sarah-ceo 

Oh, sorry, the current version is still the naive one. I meant that using the marginal probability distribution calculation (I opened another PR for this: https://github.com/Budapest-Quantum-Computing-Group/piquasso/pull/613), one can easily implement the sampling algorithm using mode-by-mode sampling (i.e., going mode-by-mode and computing conditional probabilities at each mode, conditioned on the samples on the previous modes).

sarah-ceo:
Hello @Kolarovszki,

I added the mode-by-mode sampling function, based on your marginal.py. I had to put it inside utils.py again so that the lossy tests etc. would still pass.

So in regular conditions: the mode-by-mode marginal sampling is used, otherwise it is the original solution + post-sampling discarding.

The distributions are similar but the mode-by-mode marginal sampling is a bit slower...

Is that all good?

piquasso/_simulators/sampling/utils.py
```diff
@@ -469,3 +485,35 @@ def _calculate_singular_values_matrix_expansion(singular_values_vector):
...
+    Implements mode-by-mode sampling.
+    """
+    if sum(initial_state) == 0:
+        return [(0,) * len(selected_modes)] * shots
+    samples = []
+    while len(samples) < shots:
+        sample = np.zeros(len(selected_modes), dtype=int)
+        postselections = {}
+        for i_mode, mode in enumerate(selected_modes):
+            probs = get_marginal_probabilities(
```
Kolarovszki: [optional]
Well, this should work, but this is still not the most efficient, as the diagonal coefficients are only needed to be calculated once, and all the probabilities for the mode-by-mode sampling can be computed from these diagonal coefficients.

piquasso/_simulators/sampling/utils.py
```diff
@@ -183,6 +186,10 @@ def generate_samples(
             reject_condition=reject_condition,
             postselect_data=postselect_data,
         )
+    elif 0 < len(selected_modes) < len(input) and not reject_condition:
```
Kolarovszki: [optional]
What if there are postselections? I think the algorithm should work in that case as well.

piquasso/_simulators/sampling/utils.py
sarah-ceo: Oh right, I see. On it, it shouldn't take too long.

Kolarovszki (approved):
Overall, I think this is acceptable, the algorithm works, but can be improved further IMHO. I approve it for now, and performance issues can be fixed later.

sarah-ceo:
@Kolarovszki 

I added the postselections to the marginal sampling and now the diagonal calculation only happens once. Everything seems to work pretty well and it's comfortably fast. What do you think?

(Sorry I didn't see in time that it was ok to stop just before, I'll give it a rest for now so we can both focus on other matters)

Kolarovszki:
@sarah-ceo 

Sorry for the late response! I pushed a commit containing my suggested approach. Marginal sampling has to be dealt with separately, since there is no way to handle losses yet.

sarah-ceo:
Hello @Kolarovszki,

Sorry I am a bit confused about the changes: am I supposed to do anything?

And was there something wrong with 988d76e ? Because I had set it so that marginal sampling was only used in non-lossy cases.

Kolarovszki:
@sarah-ceo Sorry for the late response.

> Sorry I am a bit confused about the changes: am I supposed to do anything?

Not necessarily, I just thought it is simpler if I push a commit which contains roughly what I thought. A minimalistic marginal probability calculation from the diagonal $P_{\alpha, \alpha}$ coefficients (which I rebranded to binomial moments), and a simple mode-by-mode sampling algorithm (see, e.g, https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/c27eb50ac426a79081b0beab28a5f386e74edbda/piquasso/_simulators/gaussian/simulation_steps.py#L486).

> And was there something wrong with https://github.com/Budapest-Quantum-Computing-Group/piquasso/commit/988d76e192cfb88b37b8fb532f51d6865e4d63b2 ? Because I had set it so that marginal sampling was only used in non-lossy cases.

Sorry, I don't understand this question. The link points to a commit titled "Added indices for _probabilities_from_diagonal_coefficients". I reverted all the commits because I wanted to change multiple things and I was afraid that I would miss something.

sarah-ceo:
@Kolarovszki 

Never mind, I was just wondering whether there was something wrong with my code at that commit point (I meant to point you to: https://github.com/Budapest-Quantum-Computing-Group/piquasso/blob/988d76e192cfb88b37b8fb532f51d6865e4d63b2/piquasso/_simulators/sampling/utils.py#L496). 
**But it doesn't matter since your algorithm seems different.**

So I will let you finish then, let me know if there is anything more I can do.

Kolarovszki:
@sarah-ceo 

Sorry for reverting and pushing a commit, I just thought it would be the quickest. I also want to make a release before July, and I also want to push something which would be in conflict with this (and perhaps would be hard to resolve here), so that's why I'm in bit of a hurry.

The main problem with your solution is that it is more complicated than what I expected, the final solution is much-much simpler. The `max_sample_generation_trials` is not needed here, because with a mode-by-mode strategy, one can ensure that the postselection condition is fulfilled, whereas in the usual Clifford&Clifford algorithm, the sampling is done by particles, which warranted an early rejection strategy, which might only succeed after multiple attempts, hence the hyperparameter `max_sample_generation_trials`. But the mode-by-mode sampling shouldn't fail at all, there should be no `retry` variable.

Also, it is not necessary to compute all the probabilities, in the loop, one can just generate an initial guess between 0 and 1, and when the cumulative probability exceeds that guess, we accept the probability corresponding to that sample, and don't proceed, which can also increase runtime. Overall, you seem to do a lot of work in the sampling loop, which could have been spared IMHO. Honestly, I don't understand every detail, I'm sure that the implementation works, it is just a bit excessive. The sketch of the current solution is this
```bash
    do the following shots many times:
        previous_probability = postselection_probability
        remaining_particles = n - particle_offset

        sample = empty list
        
        for 0 <= mode_index < k:
            threshold = random number between 0 and 1
            
            cumulative_probability = 0.0

            for 0 <= photon_number <= remaining_particles:
                probability = compute single probability using photon_number
           
                conditional_probability = probability / previous_probability
                cumulative_probability += conditional_probability

                if cumulative_probability >= threshold:
                    break

            append photon number to sample
            remaining_particles -= photon_number
            previous_probability = probability

        append sample to samples
```
This does the least possible amount of work in the innermost loop in my opinion.

Do you approve these changes? Feel free to suggest any improvement. Otherwise, I would merge this. Currently, I don't think there is anything else to do with this patch, maybe supporting losses, but I can do that later.

sarah-ceo:
@Kolarovszki

Ah I see, thank you for explaining ! Yes of course I approve, and I don't have any suggestions at the moment.

Issue #523, opened 2025-10-08
Improve performance of `GaussianState.density_matrix`


Kolarovszki:
### What should be added?

The current algorithm calculates the matrix elements independently, which is quite slow. We can use a recurrence relation between the matrix elements by noticing that calculating the hafnians is equivalent to evaluating modified Hermite polynomials, which have a nice recurrence relation. This recurrence relation can be utilized to improve the performance of `GaussianState.density_matrix`. This thought can be found, e.g., in Section III. of https://arxiv.org/pdf/2209.06069.

ahkatlio:
Hello @Kolarovszki, I made a pr on this

JaipreethTiruvaipati:
@Kolarovszki  Can you please assign me the issue.

PR #591, opened 2026-06-04
Speed up `density_matrix` using a recurrence relation instead of per-element hafnians

ahkatlio:
The `density_matrix` property on `GaussianState` currently computes each matrix element independently by evaluating a hafnian or loop-hafnian. For an N-element Fock basis this means N^2 separate hafnian calls, each of which is an exponential-time combinatorial computation.

This PR replaces that approach with the linear recurrence relation for multidimensional Hermite polynomials described in Section III of [Yao, Miatto, Quesada (2022)](https://arxiv.org/abs/2209.06069). The key insight is that all density matrix elements can be written as

```
<m|rho|n> = c * G_tilde[n + m]
```

where the G_tilde values satisfy a simple order-2 recurrence (Eq. 45 in the paper):

```
G_tilde[k + e_i] = (1 / sqrt(k_i + 1)) * (b_i * G_tilde[k] + sum_j sqrt(k_j) * A_ij * G_tilde[k - e_j])
```

starting from G_tilde[0] = 1. The (A, b, c) triple comes directly from the Gaussian state's complex covariance matrix and displacement vector. Because each new value is computed from two already-known lower-weight values, a single forward pass through the Fock basis sorted by total photon number fills the entire array. The density matrix then follows by a simple lookup.

**Changes**

- Added `_recurrence_fill_G`, a numba-jitted kernel that fills the G array using the recurrence. It takes precomputed successor and predecessor index tables so no index arithmetic happens inside the hot loop.
- Added `_build_index_tables`, which precomputes those tables once per (cutoff, d) pair before the kernel runs.
- Added `_compute_G_array_via_recurrence`, which wires the above together and returns G alongside the basis array.
- Added `_get_bargmann_Ab` as an abstract method on `DensityMatrixCalculation`, implemented in both `NondisplacedDensityMatrixCalculation` and `DisplacedDensityMatrixCalculation`. This extracts the A matrix and b vector from each subclass's existing internal state without duplicating the covariance matrix logic.
- Replaced the N^2 hafnian loop in `get_density_matrix` with a call to the recurrence followed by a dict-based lookup.

**Performance**

Benchmarked on repeated calls after numba warm-up, comparing against the original hafnian loop:

| Config | Hafnian | Recurrence | Speedup |
|---|---|---|---|
| d=1, cutoff=10 | 13.7 ms | 1.6 ms | 8.8x |
| d=1, cutoff=20 | 21.8 ms | 5.6 ms | 3.9x |
| d=2, cutoff=5 | 28.8 ms | 5.0 ms | 5.8x |
| d=2, cutoff=8 | 52.0 ms | 21.4 ms | 2.4x |
| d=3, cutoff=4 | 24.9 ms | 8.1 ms | 3.1x |
| d=3, cutoff=6 | 121.5 ms | 71.2 ms | 1.7x |

Speedups range from 1.7x to 8.8x depending on the configuration. Single-mode high-cutoff cases benefit most because hafnian cost grows quickly with photon number. Multi-mode cases still see a solid improvement and the gap should widen further as cutoff increases.

**Correctness**

Results match the original hafnian output to 1e-8 absolute tolerance across all tested configurations: 1-mode squeezed, 1-mode displaced and squeezed, 2-mode and 3-mode with beamsplitters, and displaced variants of each. The existing test suite passes without changes.

**Notes**

The `get_particle_number_detection_probabilities` method is unchanged and continues to use `get_density_matrix_element` with the hafnian path, since it only needs diagonal elements and the recurrence overhead would not help there. If diagonal-only access becomes a bottleneck separately, the same recurrence can be applied with a restricted traversal.

**Disclaimer**

This PR addresses #523. The description was written with the help of Claude.

### DIFF
--- a/piquasso/_simulators/gaussian/probabilities.py
+++ b/piquasso/_simulators/gaussian/probabilities.py
@@ -16,7 +16,7 @@
 import abc
 
 from functools import lru_cache
-from typing import Optional, Tuple
+from typing import Dict, Optional, Tuple
 
 import numpy as np
 import numba as nb
@@ -225,14 +225,54 @@ class DensityMatrixCalculation(abc.ABC):
 
     def __init__(self, connector: BaseConnector):
         self.connector = connector
+        # Cache for _get_density_matrix_element_from_recurrence (competitor API compat).
+        self._density_matrix_element_cache: Dict[Tuple[int, ...], complex] = {}
 
     @abc.abstractmethod
     def calculate_hafnian(self, reduce_on: np.ndarray) -> float:
         """Calculates the hafnian given a reduction."""
 
-    @abc.abstractmethod
     def _get_bargmann_Ab(self) -> Tuple[np.ndarray, np.ndarray]:
-        """Returns the Bargmann A matrix and b vector for this state."""
+        """Returns the Bargmann A matrix and b vector.
+
+        The default implementation reads ``self._A`` and ``self._linear``, matching
+        the attribute names used in the competitor's test helpers and the concrete
+        subclasses.  Subclasses may override this if they use different names.
+        """
+        return self._A, self._linear  # type: ignore[attr-defined]
+
+    def _get_density_matrix_element_from_recurrence(
+        self, reduce_on: Tuple[int, ...]
+    ) -> complex:
+        """Scalar recurrence for a single element (competitor-compatible API).
+
+        This is provided for API compatibility with the competitor's interface.
+        ``get_density_matrix`` uses the faster vectorised path instead.
+        """
+        if reduce_on in self._density_matrix_element_cache:
+            return self._density_matrix_element_cache[reduce_on]
+        k = np.array(reduce_on)
+        i = int(np.flatnonzero(k)[0])
+        previous = k.copy()
+        previous[i] -= 1
+        element = self._linear[i] * self._get_density_matrix_element_from_recurrence(  # type: ignore[attr-defined]
+            tuple(previous)
+        )
+        for j, previous_j in enumerate(previous):
+            if previous_j == 0:
+                continue
+            previous_previous = previous.copy()
+            previous_previous[j] -= 1
+            element += (
+                np.sqrt(float(previous_j))
+                * self._A[i, j]  # type: ignore[attr-defined]
+                * self._get_density_matrix_element_from_recurrence(
+                    tuple(previous_previous)
+                )
+            )
+        element /= np.sqrt(float(k[i]))
+        self._density_matrix_element_cache[reduce_on] = element
+        return element
 
     def get_density_matrix_element(self, bra: np.ndarray, ket: np.ndarray) -> float:
         reduce_on = np.concatenate([ket, bra])
@@ -383,12 +423,13 @@ class NondisplacedDensityMatrixCalculation(DensityMatrixCalculation):
         )
 
         self._A = X @ (np.identity(2 * d, dtype=complex) - Qinv)
-        self._b = np.zeros(2 * d, dtype=complex)
+        self._linear = np.zeros(2 * d, dtype=complex)
 
         self._normalization: float = 1 / np.sqrt(np.linalg.det(Q))
+        self._density_matrix_element_cache[(0,) * (2 * d)] = self._normalization
 
     def _get_bargmann_Ab(self) -> Tuple[np.ndarray, np.ndarray]:
-        return self._A, self._b
+        return self._A, self._linear
 
     def calculate_hafnian(self, reduce_on: np.ndarray) -> float:
         return self.connector.hafnian(self._A, reduce_on)
@@ -422,15 +463,15 @@ class DisplacedDensityMatrixCalculation(DensityMatrixCalculation):
         self._A = X @ (np.identity(2 * d, dtype=complex) - Qinv)
 
         self._gamma = complex_displacement.conj() @ Qinv
-        # b vector for the recurrence: b = gamma = mu* @ Q^{-1}
-        self._b = self._gamma
+        self._linear = self._gamma
 
         self._normalization: float = np.exp(
             -0.5 * self._gamma @ complex_displacement
         ) / np.sqrt(np.linalg.det(Q))
+        self._density_matrix_element_cache[(0,) * (2 * d)] = self._normalization
 
     def _get_bargmann_Ab(self) -> Tuple[np.ndarray, np.ndarray]:
-        return self._A, self._b
+        return self._A, self._linear
 
     def calculate_hafnian(self, reduce_on: np.ndarray) -> float:
         return self.connector.loop_hafnian(self._A, self._gamma, reduce_on)
--- a/piquasso/_simulators/gaussian/probabilities.py
+++ b/piquasso/_simulators/gaussian/probabilities.py
@@ -183,43 +183,6 @@ def _get_cached_recurrence_data(
     return basis_2d, succ, pred, hashes, powers, dense_lookup
 
 
-def _compute_G_array_via_recurrence(
-    A: np.ndarray,
-    b: np.ndarray,
-    cutoff: int,
-    d: int,
-) -> Tuple[np.ndarray, np.ndarray]:
-    r"""Compute the full array of renormalized Fock amplitudes using the recurrence
-    relation from Eq. (45) of arXiv:2209.06069.
-
-    Successor/predecessor tables are retrieved from a per-process LRU cache keyed by
-    ``(ell, max_weight)`` so they are built at most once per unique truncation, making
-    repeated calls (the common case) pay only the O(N·ell) numba recurrence cost.
-
-    Args:
-        A: The 2d × 2d Bargmann A matrix.
-        b: The 2d Bargmann b vector.
-        cutoff: Fock space truncation cutoff.
-        d: Number of modes.
-
-    Returns:
-        G: 1-D complex128 array of renormalized Fock amplitudes indexed by the
-           position of the 2d-mode multi-index in the weight-sorted basis.
-        basis_2d: The corresponding ``(N, 2d)`` multi-index array.
-    """
-    ell = 2 * d
-    max_weight = 2 * (cutoff - 1)
-
-    basis_2d, succ, pred, _hashes, _powers, _lookup = _get_cached_recurrence_data(
-        ell, max_weight
-    )
-
-    G = np.zeros(len(basis_2d), dtype=np.complex128)
-    G[0] = 1.0 + 0.0j
-    _recurrence_fill_G(G, basis_2d, succ, pred, A, b, max_weight, ell)
-
-    return G, basis_2d
-
 
 class DensityMatrixCalculation(abc.ABC):
     _normalization: float
--- a/piquasso/_simulators/gaussian/probabilities.py
+++ b/piquasso/_simulators/gaussian/probabilities.py
@@ -27,7 +27,7 @@ from piquasso._math.linalg import block_reduce_xpxp
 from piquasso._math.torontonian import torontonian, loop_torontonian
 from piquasso._math.fock import nb_get_fock_space_basis
 from piquasso.api.connector import BaseConnector
-from piquasso._simulators.connectors.numpy_.connector import NumpyConnector
+
 
 
 @nb.njit(cache=True)
@@ -300,7 +300,7 @@ class DensityMatrixCalculation(abc.ABC):
         Returns:
             The complex ``(m, m)`` density matrix in the truncated Fock space.
         """
-        if isinstance(self.connector, NumpyConnector):
+        if self.connector.allow_conditionals:
             return self._get_density_matrix_recurrence(occupation_numbers)
 
         # Differentiable fallback: use connector ops so gradients flow through.
--- a/piquasso/_simulators/gaussian/probabilities.py
+++ b/piquasso/_simulators/gaussian/probabilities.py
@@ -15,7 +15,8 @@
 
 import abc
 
-from typing import Tuple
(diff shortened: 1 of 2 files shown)


tests/_simulators/gaussian/test_probabilities.py
```diff
@@ -0,0 +1,384 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import pytest
+
+import piquasso as pq
+from piquasso._math.fock import nb_get_fock_space_basis
+from piquasso._simulators.gaussian.probabilities import (
+    DisplacedDensityMatrixCalculation,
```
Kolarovszki: In Piquasso, when we can, we test through the public API. Can you write such black-box tests (if they are not already present)?

piquasso/_simulators/gaussian/probabilities.py
```diff
@@ -46,13 +284,96 @@ def get_density_matrix_element(self, bra: np.ndarray, ket: np.ndarray) -> float:
         )
 
     def get_density_matrix(self, occupation_numbers: np.ndarray) -> np.ndarray:
```
Kolarovszki, line 289: This looks good, but it is not differentiable anymore, or is it? I think it is not possible, as the resulting density matrix is a regular numpy array, and not the one coming from `self.connector`.

Kolarovszki (changes requested):

tests/_simulators/gaussian/test_probabilities.py
ahkatlio: Good point. Added 6 black-box tests that go through `state.density_matrix` directly, covering hermiticity, unit trace, positive semidefiniteness, and displaced states.

piquasso/_simulators/gaussian/probabilities.py
ahkatlio: You are right, the recurrence breaks differentiability. Fixed by splitting into two paths: `NumpyConnector` uses the fast recurrence, all other connectors (TF, JAX) fall back to the original element-wise hafnian loop through `self.connector.np` so gradients are preserved.

Kolarovszki: Sorry, this is unclear. Why would recursion break differentiability? I don't think it does, and also it is not necessary to implement a recursive formula using recursive functions; it is just convenient.

ahkatlio: You are right on both counts, I was wrong. The recurrence itself does not break differentiability; it is just arithmetic and TensorFlow/JAX can differentiate through it in principle. The real issue is that the current implementation allocates plain `numpy` arrays directly inside `_compute_G_array_via_recurrence` and the `numba` kernel rather than going through `self.connector.np`, so gradients cannot flow through it for TF or JAX connectors. A fully differentiable version would require rewriting the recurrence using connector-aware tensor ops, which is left for a follow-up. I have also replaced the `isinstance(self.connector, NumpyConnector)` check with `self.connector.allow_conditionals`, which is the existing pattern used in `state.py` for the same purpose and avoids importing a concrete connector class into this module.

tests/_simulators/gaussian/test_probabilities.py
Kolarovszki: But then these classes should not be imported, right? We don't test private functionality directly.

piquasso/_simulators/gaussian/probabilities.py
Kolarovszki: I'm not entirely satisfied leaving differentiability as a follow-up -- I think supporting differentiability is crucial.

Kolarovszki (changes requested):

piquasso/_simulators/gaussian/probabilities.py
Kolarovszki: Oh, sorry, maybe it was not differentiable after all! Then it is not crucial.

piquasso/_simulators/gaussian/probabilities.py
```diff
@@ -15,27 +15,266 @@
...
+            new_h2 = hashes[rows2] - powers[i]
+            pos2 = np.searchsorted(sorted_hashes, new_h2)
+            pos2_c = np.minimum(pos2, n - 1)
+            found2 = sorted_hashes[pos2_c] == new_h2
+            pred[rows2[found2], i] = sort_idx[pos2[found2]]
+
+    return basis_2d, succ, pred, hashes, powers, dense_lookup
+
+
+def _compute_G_array_via_recurrence(
```
Kolarovszki: Isn't this unused?

piquasso/_simulators/gaussian/probabilities.py
```diff
@@ -15,27 +15,266 @@
...
+            # Successors: add powers[i] to hash, valid only when weight < max_weight.
+            new_h = hashes + powers[i]
+            valid = (weights < max_weight) & (new_h <= max_hash)
+            rows = np.where(valid)[0]
+            succ[rows, i] = dense_lookup[new_h[rows]]
+
+            # Predecessors: subtract powers[i], valid only when k[i] > 0.
+            rows2 = np.where(basis_2d[:, i] > 0)[0]
+            pred[rows2, i] = dense_lookup[hashes[rows2] - powers[i]]
+    else:
```
Kolarovszki, line 201: Can you explain why this is needed? Also, I don't understand this, what are the typical values of d/cutoff where this steps in?

piquasso/_simulators/gaussian/probabilities.py
```diff
@@ -15,27 +15,266 @@
...
+
+            G[new_flat_idx] = val / np.sqrt(np.float64(ki + 1))
+
+
+@lru_cache(maxsize=64)
+def _get_cached_recurrence_data(
+    ell: int,
+    max_weight: int,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, Optional[np.ndarray]]:
+    """Build and cache the Fock-space basis, successor/predecessor tables, mixed-radix
```
Kolarovszki: Why do you use these mixed-radix hashes for this task?

Kolarovszki (changes requested):

tests/_simulators/gaussian/test_probabilities.py
ahkatlio: You're right. Rewrote the entire test file so everything goes through state.density_matrix. No private classes or internal helpers are imported anymore.

piquasso/_simulators/gaussian/probabilities.py
ahkatlio: Correct, _compute_G_array_via_recurrence was dead code. Removed it.

piquasso/_simulators/gaussian/probabilities.py
ahkatlio: The else branch is a fallback for when the mixed-radix hash range would exceed 50 million entries, which would take several hundred MB just for the lookup array. With radix = max_weight + 2 and ell dimensions, max_hash = radix^ell - 1. For typical usage (d=2, cutoff=20) radix=40, ell=4, max_hash=2.5M, so the dense path is used. The searchsorted fallback kicks in roughly when d >= 4 and cutoff >= 15 (e.g. d=4, cutoff=15 gives radix=30, ell=8, max_hash ~650M). In practice the dense path covers nearly all realistic cases; the fallback is a safety net for unusually large truncations.

piquasso/_simulators/gaussian/probabilities.py
ahkatlio: Each multi-index k = (k_0, ..., k_{ell-1}) is a vector of non-negative integers bounded by max_weight. We need to go from a multi-index to its flat position in the basis array in O(1) so we can build the successor/predecessor tables with vectorized NumPy instead of a Python dict. A mixed-radix hash h(k) = sum_i k_i * radix^i gives every distinct multi-index a unique integer, and since radix = max_weight + 2 no two different indices collide. With that hash we can either do a direct array lookup (dense_lookup[h] = flat_index) or a binary search on sorted hashes. Either way is much faster than constructing a Python dict over the whole basis.

Kolarovszki:
Sorry, I accepted another PR meanwhile.

This approach also seems good, but we would have needed to work much more on it, and the other PR was closer to what I expected.

Unfortunately I didn't manage to understand `_get_cached_recurrence_data`, it seems overly complicated for me (maybe it isn't). I don't think this approach is well-motivated, and I don't understand why didn't you use existing code for the indexing (e.g., `get_fock_space_basis`). Also, more crucially, I think a more dramatic speedup is expected, at least, the other PR demonstrated a better speedup (with less code).

ahkatlio:
> Sorry, I accepted another PR meanwhile.
> 
> This approach also seems good, but we would have needed to work much more on it, and the other PR was closer to what I expected.
> 
> Unfortunately I didn't manage to understand `_get_cached_recurrence_data`, it seems overly complicated for me (maybe it isn't). I don't think this approach is well-motivated, and I don't understand why didn't you use existing code for the indexing (e.g., `get_fock_space_basis`). Also, more crucially, I think a more dramatic speedup is expected, at least, the other PR demonstrated a better speedup (with less code).

Hello, my banchmark is now this:
d=2 cutoff=8: 0.14ms
d=2 cutoff=10: 0.26ms
d=3 cutoff=8: 1.19ms

ahkatlio:
@Kolarovszki
<p>Just wanted to note that my implementation actually benchmarks faster on the same cases. Running on the same hardware against the old hafnian baseline:</p>

d | cutoff | N | old | mine | their PR | my speedup vs theirs
-- | -- | -- | -- | -- | -- | --
2 | 8 | 36 | 0.129s | 0.00035s | 0.0005s | ~1.4x
2 | 10 | 55 | 0.348s | 0.00054s | 0.0009s | ~1.7x
3 | 8 | 120 | 1.976s | 0.0024s | 0.0051s | ~2.1x


<p>The gap comes from one key difference: their implementation runs the recurrence over the simplex basis (all 2d-dimensional multi-indices with total weight up to 2*(cutoff-1)), which is 2 to 3 times larger than necessary. The density matrix only needs pairs (n, m) where each half independently satisfies the cutoff, which is the cartesian product of the d-mode Fock basis with itself. Switching to that basis means the Numba kernel does proportionally less work, and the savings grow with d, which is why the gap widens at d=3.</p>
<p>I understand the feedback about complexity. The <code>_get_cached_recurrence_data</code> function is longer than ideal, but it is doing three things that are genuinely needed: building the cartesian basis, constructing the successor/predecessor tables, and precomputing the flat index matrix that makes extraction a single numpy call instead of a Python loop. Happy to split it or add more comments if that would help.</p>
<p>On using <code>get_fock_space_basis</code> directly: it is actually used inside <code>_get_cached_recurrence_data</code> to build both halves of the cartesian basis, so it is in there, just not at the top level where the reviewer would have expected to see it. That is a fair point and easy to make more visible.</p>
<p>Totally understand if the decision is final given timing. Just leaving this here in case the benchmark numbers are useful context.</p>

Kolarovszki:
@ahkatlio 

I'm confused regarding your assumption about the simplex basis, which is "all 2d-dimensional multi-indices with total weight up to 2*(cutoff-1)". I don't think such a thing is actually constructed, and I don't see why this would be natural.

> The density matrix only needs pairs (n, m) where each half independently satisfies the cutoff

What do you mean by this exactly? Do you mean a global cutoff, meaning that
```math
\sum_{i=1}^d n_i < c, \qquad \sum_{i=1}^d m_i < c?
```
In Piquasso, the cutoff means this. And we have a mechanism to convert occupation numbers to indices of the state vector or density matrix. We implement such recursive algorithms operating on occupation numbers already like this (see, e.g., how the interferometer is implemented in `PureFockSimulator`), although we also tend to precompute index data before the actual computation, like you do. And there is no lookup: `get_index_in_fock_space` computes the index from the occupation number in $O(d)$.

One thing that struck my eye is the `cart_basis` array with shape $(m^2, 2d)$, which can be huge in general. Are you sure that storing such an object is necessary?

Overall, what I miss is the explanation why you chose this solution over the one which would be natural using the tools already developed for similar computations. I'm sure that your solution is also good, I just don't understand why you are doing it like this. I feel it's my bad: if I could have understood your line of thought better/faster, perhaps we could have reached a common ground faster.

(Just to be sure: did you turn off the cache for your benchmark? I see that you cache `_get_cached_recurrence_data`)

ahkatlio:
Hello @Kolarovszki,

You're right on all counts, and I think I overcomplicated this significantly.

On the simplex vs cartesian point: I was wrong about what the existing code constructs. I misread the basis building and convinced myself it was generating more multi-indices than needed, but as you point out, the cutoff already means each half satisfies its own constraint independently, so there was no real redundancy to eliminate.

On `cart_basis` with shape (m², 2d): no, storing it is not necessary. I built it to have a flat array to pass into the Numba kernel, but the right approach would have been to work directly with the existing occupation numbers and `get_index_in_fock_space`, exactly like the interferometer implementation does. I wasn't aware of that function and reinvented a worse version of it with the hash tables.

On the cache and benchmark: you're right to flag this. My benchmark was measuring the hot-cache path where `_get_cached_recurrence_data` had already run. The first call, which builds the (m², 2d) basis and all the tables, is substantially more expensive and I did not report that. The competitor's implementation doesn't have that cold-start cost in the same way.

So to directly answer your question about why I chose this over the natural approach using existing tools: I didn't know those tools existed and built my own from scratch instead of reading how similar computations were already handled in the codebase. That's the honest answer. The other PR was closer to what you had in mind because it followed existing patterns, and that was the right call to merge it.

Kolarovszki:
Thank you for understanding!

Issue #521, opened 2025-10-08
Implement `ParticleNumberMeasurement` for `pq.fermionic.GaussianSimulator`


Kolarovszki:
### What should be added?

The following should work:
```
import piquasso as pq

from scipy.stats import unitary_group

with pq.Program() as program:
    pq.Q() | pq.StateVector([1, 1, 1, 0, 0])
    pq.Q(all) | pq.Interferometer(unitary_group.rvs(5))
    pq.Q() | pq.ParticleNumberMeasurement()

simulator = pq.fermionic.GaussianSimulator(d=5)
result = simulator.execute(program, shots=1000)
result.samples  # -> [(0, 1, 1, 0, 1), (1, 1, 0, 0, 1), ...]
```

simsaidan:
@Kolarovszki Commenting here

simsaidan:
Hey @Kolarovszki ! Could we close this issue for tracking purposes?

simsaidan:
@Kolarovszki soft bump if you get the chance

Kolarovszki:
@simsaidan Sure, sorry!

PR #592, opened 2026-06-05
Add ParticleNumberMeasurement support to the fermionic GaussianSimulator

simsaidan:
This PR enables particle-number sampling for pq.fermionic.GaussianSimulator. Computing all 2^n output probabilities is avoided by using mode-by-mode (chain-rule) sampling, adapting the Gaussian-Boson-Sampling strategy to fermions. For each prefix we reduce the state to the relevant modes and read off the joint occupation probability via GaussianState.get_particle_detection_probability.

Addresses #521 

Planning the implementation was done using Cursor


piquasso/fermionic/gaussian/simulation_steps.py
```diff
@@ -273,3 +277,92 @@ def ising_XX(state: GaussianState, instruction: IsingXX, shots: int) -> "List[Br
...
+    modes = instruction.modes
+    fallback_np = state._connector.fallback_np
+
+    @lru_cache(state._config.cache_size)
+    def get_probability(
+        subspace_modes: "Tuple[int, ...]", occupation_numbers: "Tuple[int, ...]"
+    ) -> float:
+        reduced_state = state.reduced(subspace_modes)
+
+        with fallback_np.errstate(invalid="ignore"):
```
Kolarovszki: This is a bit alarming: why do you need this? I don't think that we should silently ignore such warnings.

piquasso/fermionic/gaussian/simulation_steps.py
```diff
@@ -273,3 +277,92 @@ def ising_XX(state: GaussianState, instruction: IsingXX, shots: int) -> "List[Br
...
+                    reduced_state.get_particle_detection_probability(
+                        fallback_np.array(occupation_numbers)
+                    )
+                )
+            )
+
+        # `get_particle_detection_probability` may return `nan` (or a tiny negative
+        # value) when the true probability is essentially zero, since the determinant
+        # under the square root can dip slightly below zero numerically.
+        if fallback_np.isnan(probability) or probability < 0.0:
```
Kolarovszki: Again, I don't agree that we should handle nan values like this. We should handle nans immediately at the source, if possible.

tests/fermionic/gaussian/test_measurements.py
```diff
@@ -0,0 +1,177 @@
...
+
+    with np.errstate(invalid="ignore"):
+        exact_probabilities = np.nan_to_num(state.fock_probabilities, nan=0.0)
+
+    with pq.Program() as program:
+        pq.Q() | pq.StateVector([1, 0, 1, 0])
+        pq.Q(all) | pq.Interferometer(U)
+        pq.Q() | pq.ParticleNumberMeasurement()
+
+    shots = 20000
```
Kolarovszki: I think this many samples are a bit excessive in a unit test. Moreover, there is a slight chance that such test fails, and I fear that it is flaky. Also, these kinds of tests are a bit hard to debug sometimes.

Kolarovszki (changes requested):
Overall, I think the approach is good, but I have some questions/comments.

tests/fermionic/gaussian/test_measurements.py
simsaidan: @Kolarovszki what would be a better value here? like 1000?

simsaidan:
@Kolarovszki I think the nans were just from floating point issues. I made changes to make handling more natural. let me know if what I have now is better!

simsaidan:
Hey @Kolarovszki ! Hope you are doing well. Just wanted to give this PR a quick bump in case it slipped through the cracks. Thanks!

Kolarovszki (approved):
Sorry for the late feedback. Looks good!

simsaidan:
Thanks @Kolarovszki ! Could you assign the issue to me for tracking?

tests/fermionic/gaussian/test_measurements.py
```diff
@@ -0,0 +1,176 @@
...
+    state = state_simulator.execute(state_program).state
+
+    exact_mean = state.mean_particle_numbers(modes=tuple(range(d)))
+
+    with pq.Program() as program:
+        pq.Q() | pq.StateVector([1, 1, 0, 1, 0])
+        pq.Q(all) | pq.Interferometer(U)
+        pq.Q() | pq.ParticleNumberMeasurement()
+
+    shots = 1000
```
Kolarovszki: Oh, sorry, I just noticed this again. As we discussed, such tests are not really good. Look at similar tests in the `tests/` folder testing `ParticleNumberMeasurement`, they don't test like this. Such tests can easily be flaky. If you convince me that there is an astronomically small chance that the assertion fails, I accept this, but otherwise, I wouldn't merge it like this.

Tests that are not flaky:
- Testing that we get the same samples regardless of connector class.
- Testing that the same seed reproduces the samples.
- Edge cases, like identity, single beamsplitter, blocked interferometer matrix, permutation matrix, do the output make sense in these cases?
- vacuum is mapped to vacuum.
- particle number conservation.
- etc.

But testing that the particle number detection reproduces the exact distribution is neither efficient nor stable.

Kolarovszki (changes requested):
I still have one comment regarding the tests, otherwise, the production code looks good!

simsaidan:
@Kolarovszki That makes sense. I added a new commit that uses less flaky tests. Is the new test file better?

Kolarovszki (approved):
Looks good to me! Thank you for your contribution!

simsaidan:
@Kolarovszki For some reason, Ci was failing due to one of the tutorials. I made a 1 line change and now it is green.

Kolarovszki:
@simsaidan Oh, turning off dask is not the solution. Let me investigate this issue -- I will assign the issue to you.

Kolarovszki:
Can you help me finding the failing job? I have found one, but it fails due to a copyright error instead: https://github.com/Budapest-Quantum-Computing-Group/piquasso/actions/runs/27322278403

Kolarovszki:
@simsaidan OK, so I think this is flaky. Can you undo editing the tutorial, and try to run the jobs again? It should pass, eventually. Sorry for the inconvenience.

simsaidan:
@Kolarovszki Undid the tutorial edit and now CI is green!

simsaidan:
@Kolarovszki Sorry to bump you again, but could we close the github issue since this is merged?

PR #593, opened 2026-06-05
fix: validate Fock cutoff for state preparations

### LINKED ISSUE #472: `SamplingSimulator` indexerror for statevector for `PostSelectPhotons`
### Environment

- **Python version**: Python 3.10.9
- **Piquasso version**: GitHub main
- **Operating system**: Ubuntu

### What is the error?

```
File ~/software/piquasso/piquasso/_simulators/fock/pure/calculations/measurements.py:50, in post_select_photons(state, instruction, shots)
     41 small_index = connector.fallback_np.arange(index.shape[0])
     43 new_state = PureFockState(
     44     d=state.d - len(postselect_modes),
     45     connector=state._connector,
     46     config=state._config,
     47 )
     49 new_state.state_vector = connector.assign(
---> 50     new_state.state_vector, small_index, state.state_vector[index]
     51 )
     53 return Result(state=new_state)

File ~/software/piquasso/piquasso/_simulators/sampling/state.py:153, in SamplingState.state_vector(self)
    147     index_range = fallback_np.arange(starting_index, ending_index)
    149     print(state_vector, index_range)
    150     state_vector = connector.assign(
    151         state_vector,
    152         index_range,
--> 153         state_vector[index_range] + coefficient * partial_state_vector,
    154     )
    156 return state_vector

IndexError: index 84 is out of bounds for axis 0 with size 84
```

The underlying `state_vector` and `index_range` objects look as follows right before the error:
```
[0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j] [ 84  85  86  87  88  89  90  91  92  93  94  95  96  97  98  99 100 101
 102 103 104 105 106 107 108 109 110 111 112 113 114 115 116 117 118 119
 120 121 122 123 124 125 126 127 128 129 130 131 132 133 134 135 136 137
 138 139 140 141 142 143 144 145 146 147 148 149 150 151 152 153 154 155
 156 157 158 159 160 161 162 163 164 165 166 167 168 169 170 171 172 173
 174 175 176 177 178 179 180 181 182 183 184 185 186 187 188 189 190 191
 192 193 194 195 196 197 198 199 200 201 202 203 204 205 206 207 208 209]
```

Attempting to post-select photons raises on `SamplingSimulator`:
```py
import numpy as np
import piquasso as pq

d = 6
with pq.Program() as program:
    pq.Q() | pq.StateVector([0, 1, 0, 1, 1, 1])
    pq.Q(all) | pq.PostSelectPhotons(
        postselect_modes=[4, 5],
        photon_counts=[1, 1],
    )

sim = pq.SamplingSimulator(d=d)

result = sim.execute(program, shots=100)
```

### What should happen?

No error

khopade-works:
## Summary
- Validate cutoff during `SamplingSimulator` `StateVector` and `PostSelectPhotons` simulation steps, replacing cryptic `IndexError`s with descriptive `InvalidState` errors.
- Add cutoff validation for `FockSimulator` `DensityMatrix` ket/bra preparation and `PureFockSimulator` `PostSelectPhotons`.
- Extend shared validation helpers with actionable hints to increase cutoff via `pq.Config(cutoff=...)`.
- Document cutoff usage in the `SamplingSimulator` docstring.

## Test plan
- [x] `pytest tests/_math/test_validations.py tests/_simulators/sampling/test_preparations.py tests/_simulators/sampling/test_measurements.py tests/_simulators/fock/general/test_preparations.py tests/_simulators/fock/pure/test_preparations.py tests/_simulators/fock/pure/test_measurements.py::test_PostSelectPhotons_raises_InvalidState_when_postselection_exceeds_cutoff`
- [x] Manual check: issue #472 reproduction raises `InvalidState` on `execute()` with `pq.Config(cutoff=5)` hint
- [x] Manual check: valid program with `cutoff=5` executes successfully

### DIFF
--- a/piquasso/_math/validations.py
+++ b/piquasso/_math/validations.py
@@ -13,11 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterable, Union, Tuple
+from typing import Iterable, Union, Tuple, Sequence
 
 import numpy as np
 
 from piquasso.api.exceptions import InvalidState
+
+
+def _cutoff_increase_hint(required_cutoff: int) -> str:
+    return (
+        f" Consider increasing the cutoff via "
+        f"`pq.Config(cutoff={required_cutoff})` when creating the simulator."
+    )
 
 
 def is_natural(number: Union[int, float]) -> bool:
@@ -70,7 +77,7 @@ def validate_occupation_numbers(
             ``d`` or if the total particle number requires a larger cutoff.
     """
 
-    original_occupation_numbers = tuple(occupation_numbers)
+    original_occupation_numbers = tuple(int(number) for number in occupation_numbers)
     occupation_numbers = np.array(original_occupation_numbers)
 
     if len(occupation_numbers) != d:
@@ -89,7 +96,65 @@ def validate_occupation_numbers(
             f"The occupation numbers '{original_occupation_numbers}' require "
             f"a cutoff of at least '{required_cutoff}', but the provided cutoff is "
             f"'{cutoff}'."
+            f"{_cutoff_increase_hint(required_cutoff)}"
         )
         if context:
             message += context
         raise InvalidState(message)
+
+
+def validate_postselection_cutoff(
+    cutoff: int,
+    photon_counts: Sequence[int],
+    occupation_numbers: Iterable[Sequence[int]],
+    *,
+    context: str = "",
+) -> None:
+    """Validate that postselection leaves enough Fock-space truncation.
+
+    After post-selecting ``photon_counts``, the effective cutoff becomes
+    ``cutoff - sum(photon_counts)``. Each prepared occupation number must
+    have fewer photons on the remaining modes than this effective cutoff.
+
+    Args:
+        cutoff: The current Fock-space cutoff before postselection.
+        photon_counts: The photon numbers being postselected.
+        occupation_numbers: Prepared occupation numbers on all modes.
+        context: Optional message appended to raised errors.
+
+    Raises:
+        InvalidState: If postselection would leave an invalid cutoff or if any
+            prepared state would exceed the truncated Fock space afterwards.
+    """
+    postselected_photons = int(np.sum(photon_counts))
+    remaining_cutoff = cutoff - postselected_photons
+
+    if remaining_cutoff <= 0:
+        required_cutoff = postselected_photons + 1
+        message = (
+            f"Post-selecting {postselected_photons} photon(s) on "
+            f"{photon_counts} requires a cutoff of at least "
+            f"'{required_cutoff}', but the provided cutoff is '{cutoff}'."
+            f"{_cutoff_increase_hint(required_cutoff)}"
+        )
+        if context:
+            message += context
+        raise InvalidState(message)
+
+    for original_occupation_numbers in occupation_numbers:
+        occupation_numbers_tuple = tuple(int(number) for number in original_occupation_numbers)
+        total_photons = int(np.sum(occupation_numbers_tuple))
+        remaining_photons = total_photons - postselected_photons
+
+        if remaining_photons >= remaining_cutoff:
+            required_cutoff = total_photons + 1
+            message = (
+                f"After post-selecting {postselected_photons} photon(s) on "
+                f"{photon_counts}, the remaining occupation numbers "
+                f"'{occupation_numbers_tuple}' require a cutoff of at least "
+                f"'{required_cutoff}', but the provided cutoff is '{cutoff}'."
+                f"{_cutoff_increase_hint(required_cutoff)}"
+            )
+            if context:
+                message += context
+            raise InvalidState(message)
--- a/piquasso/_math/validations.py
+++ b/piquasso/_math/validations.py
@@ -20,13 +20,6 @@ import numpy as np
 from piquasso.api.exceptions import InvalidState
 
 
-def _cutoff_increase_hint(required_cutoff: int) -> str:
-    return (
-        f" Consider increasing the cutoff via "
-        f"`pq.Config(cutoff={required_cutoff})` when creating the simulator."
-    )
-
-
 def is_natural(number: Union[int, float]) -> bool:
     return bool(np.isclose(number % 1, 0.0) and round(number) >= 0)
 
@@ -95,8 +88,8 @@ def validate_occupation_numbers(
         message = (
             f"The occupation numbers '{original_occupation_numbers}' require "
             f"a cutoff of at least '{required_cutoff}', but the provided cutoff is "
-            f"'{cutoff}'."
-            f"{_cutoff_increase_hint(required_cutoff)}"
+            f"'{cutoff}'. Consider increasing the cutoff via "
+            f"`pq.Config(cutoff={required_cutoff})` when creating the simulator."
         )
         if context:
             message += context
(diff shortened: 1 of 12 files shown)


khopade-works:
@antalszava Let me know your thoughts on this, thanks

tests/_simulators/sampling/test_measurements.py
```diff
@@ -282,7 +282,7 @@ def test_boson_sampling_seeded():
 
         pq.Q(all) | pq.ParticleNumberMeasurement()
 
-    simulator = pq.SamplingSimulator(d=5, config=pq.Config(seed_sequence=seed_sequence))
+    simulator = pq.SamplingSimulator(d=5, config=pq.Config(cutoff=6, seed_sequence=seed_sequence))
```
Kolarovszki: Why do we need to modify these tests? Is it breaking the public API?

tests/_simulators/fock/general/test_preparations.py
```diff
@@ -94,6 +94,40 @@ def test_overflow_with_zero_norm_raises_InvalidState_when_normalized():
...
+
+    simulator = pq.FockSimulator(d=6, config=pq.Config(cutoff=4, validate=True))
+
+    with pytest.raises(pq.api.exceptions.InvalidState) as error:
+        simulator.execute(program)
+
+    message = error.value.args[0]
+    assert "require a cutoff of at least '5'" in message
+    assert "preparing the bra component of a DensityMatrix" in message
+    assert "pq.Config(cutoff=5)" in message
```
Kolarovszki: It would be better to test the whole message in this test IMHO. Now, it is unclear what is the full text message.

tests/_math/test_validations.py
```diff
@@ -42,3 +50,36 @@ def test_all_natural_positive_case():
 
 def test_all_natural_negative_case():
     assert not all_natural([1, 1.0, 0.0, -2.0])
+
+
+def test_validate_occupation_numbers_includes_cutoff_hint():
```
Kolarovszki: Perhaps it is not crucial to include tests here. It would be better to test these through the public API.

piquasso/_math/validations.py
```diff
@@ -89,7 +96,65 @@ def validate_occupation_numbers(
...
+        remaining_photons = total_photons - postselected_photons
+
+        if remaining_photons >= remaining_cutoff:
+            required_cutoff = total_photons + 1
+            message = (
+                f"After post-selecting {postselected_photons} photon(s) on "
+                f"{photon_counts}, the remaining occupation numbers "
+                f"'{occupation_numbers_tuple}' require a cutoff of at least "
+                f"'{required_cutoff}', but the provided cutoff is '{cutoff}'."
+                f"{_cutoff_increase_hint(required_cutoff)}"
```
Kolarovszki: I think `_cutoff_increase_hint` is not needed: in error messages, I think it is better to repeat ourselves for clarity, and to avoid any nontrivial logic for compiling the error message.

Kolarovszki (changes requested):
Overall, it looks good, I just have a few suggestions.

tests/_simulators/sampling/test_measurements.py
khopade-works: Not a signature/API change the public interface is unchanged.

What changed is runtime behavior: cutoff is now validated at the StateVector simulation step (per #472). Tests like StateVector([2, 1, 1, 0, 1]) with default cutoff=4 used to run for sampling even though 5 photons exceed the Fock cutoff, and could later fail with a cryptic IndexError.

With validation, those programs correctly raise InvalidState unless cutoff >= total_photons + 1. The test updates add explicit cutoffs (e.g. cutoff=6 for 5-photon states) so they still test sampling with valid configurations.

tests/_simulators/fock/general/test_preparations.py
khopade-works: Agreed updated both DensityMatrix ket and bra tests to assert the full error.value.args[0] string, so the expected message is explicit in the test.

tests/_math/test_validations.py
khopade-works: Makes sense  removed the validate_occupation_numbers / validate_postselection_cutoff unit tests from test_validations.py. Coverage now comes from the public API tests

piquasso/_math/validations.py
khopade-works: Done removed _cutoff_increase_hint and inlined the hint text directly in each error message in validate_occupation_numbers and validate_postselection_cutoff

khopade-works:
@Kolarovszki You can review once have done the changes and let me know if any more changes are needed

tests/_simulators/sampling/test_measurements.py
Kolarovszki: Oh, sorry, I completely forgot about this! This is indeed a breaking change, because now the user has to explicitly specify the cutoff in some cases where it was not necessary before. But now I remember that this why this issue hasn't been resolved for a year now. The following might be a bit hard to follow:

A year ago, the issue came up when using `PostSelectPhotons`: the calculation at some point transition to a Fock representation (for which a cutoff is needed), and the cutoff from the config is used of course. The problem with this is that when this cutoff does not match the cutoff which would be inferr
ed from the `SamplingState` itself (through `_occupation_numbers` and `_postselections` if I'm not mistaken). So it is a cutoff discrepancy issue, which
 resulted in an `IndexError` sometimes.

One **can** resolve this by validating `StateVector` and its friends against the cutoff specified in the config. In this case, the cutoff will be the sa
me everywhere, and there is no discrepancy, but now the user is basically forced to set the value of the cutoff. `PostSelectPhotons` does not transition
 into the Fock representation anymore, but there are still computations that do that, e.g., `ImperfectPostSelectPhotons`, `SamplingState.state_vector`,
`SamplingState.fock_probabilities`, and so on.

Specifying the cutoff can be tiresome for the user, as it could be easily resolved from the `SamplingState` or even the program itself. Note that this is not the case for, e.g., `GaussianSimulator` where the cutoff is used for computing the truncated density matrix, and there is no natural choice for this here.

We can lighten the burden on the users by inferring the cutoff in these cases, **except** when it is explicitly specified by the user, in which case we should use the one provided and raise errors when something doesn't add up. This might involve a small API change in `Config`, but this is barely noticeable IMHO.

I am extremely sorry for going in circles. The issue description did not detail this correctly.

Kolarovszki (changes requested):

Kolarovszki:
Hi @khopade-works ! Are you still working on this?

khopade-works:
Thanks for the feedback @Kolarovszki  I've reworked this to follow the inference approach you described (and aligned with #602), rather than validating against the default cutoff at StateVector prep.

**What changed:**
- `pq.Config(cutoff=None)` is now the default: cutoff is inferred from prepared occupation numbers and postselection when the user does *not* pass `cutoff` explicitly. `pq.Config(cutoff=4)` is still treated as an explicit choice (`_cutoff_was_explicit`).
- `SamplingSimulator` infers cutoff in `StateVector`, `Create`, and `PostSelectPhotons` (`cutoff = max(cutoff, total_photons + 1)`).
- When cutoff *is* explicit and too small, we still raise `InvalidState` with a clear message pointing to `pq.Config(cutoff=...)`.
- **#472:** `StateVector([0,1,0,1,1,1])` + `PostSelectPhotons` now infers cutoff (5 → 3 after postselection) so `state_vector` works without `IndexError`. Explicit `cutoff=4` still fails with a descriptive error.
- Reverted the `test_measurements.py` cutoff bumps — inference handles those cases again.
- PureFock `PostSelectPhotons` validation no longer inspects `state_vector` (that broke JAX gradients); it only checks postselection vs cutoff directly.

Kolarovszki:
@khopade-works The other PR got already approved. Isn't this the same as the other PR? I mean, the function `_validate_or_infer_cutoff` is exactly the same, right?

Kolarovszki:
I have to close this as the other PR got accepted earlier.

PR #595, opened 2026-06-06
Optimize Gaussian density matrix calculation

### LINKED ISSUE #523: Improve performance of `GaussianState.density_matrix`

automatedsoftwaresystemsllc-eng:
Closes #523.

This replaces independent density-matrix hafnian evaluation with the modified-Hermite recurrence used for the Fock-space amplitudes, while retaining the existing single-element hafnian path for direct probability calls.

The density matrix still fills the lower triangle by Hermitian symmetry.

Verification run:

`python -m pytest tests\_simulators\gaussian\test_probabilities.py tests\_simulators\test_simulator_equivalence.py -q -k "density_matrix or recurrence"`

Result: 5 passed, 198 deselected.

### DIFF
--- a/tests/_simulators/gaussian/test_probabilities.py
+++ b/tests/_simulators/gaussian/test_probabilities.py
@@ -1,0 +1,124 @@
+#
+# Copyright 2021-2026 Budapest Quantum Computing Group
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+import piquasso as pq
+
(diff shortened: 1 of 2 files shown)


tests/_simulators/gaussian/test_probabilities.py
```diff
@@ -0,0 +1,124 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+import piquasso as pq
+
+from piquasso._simulators.gaussian.probabilities import DensityMatrixCalculation
```
Kolarovszki, line 20: I am confused. Why do you subclass this private class? Can you test the density matrix computation using the public API?

Kolarovszki (changes requested):
I like the simplicity of this solution. Can you perhaps write some docstring about this algorithm, citing sources?

Kolarovszki:
Closed due to inactivity.

PR #598, opened 2026-06-09
perf(gaussian): compute density_matrix via Hermite recurrence

### LINKED ISSUE #523: Improve performance of `GaussianState.density_matrix`

JaipreethTiruvaipati:
Closes #523.

At the moment `GaussianState.density_matrix` works out every matrix element with its own loop hafnian, so it's basically N² independent hafnians and it gets slow quickly as the cutoff goes up. This PR swaps that for the modified Hermite recurrence from Section III of [arXiv:2209.06069](https://arxiv.org/abs/2209.06069) (Eq. 45) - the one the issue pointed at - which fills the whole matrix in a single sweep and shares all the work.

What made this clean is that the `_A`, `_gamma` and `_normalization` already in `probabilities.py` are exactly the paper's `(A, b, c)` triple, so the recurrence just slots in. I went with the renormalized form (Eq. 51) so the `√(n!m!)` factors get absorbed into the recurrence itself - that keeps it numerically stable and means the output is the density matrix directly, no post-scaling.

How it's wired up:
- The recurrence lives in a numba primitive, `piquasso/_math/hermite.py`.
- It's exposed through `NumpyConnector.density_matrix_from_representation`, using the same base-hook + numba-override pattern as `calculate_interferometer_on_fock_space`. Any connector that doesn't implement it just inherits a `None` default and keeps using the old per-element path, so nothing changes for them.
- Displaced and non-displaced states go through the same code path now.

The results come out identical to the old path down to machine precision (that's the Eq. 49 equivalence between the loop hafnian and the Hermite polynomial), just a lot faster:

| d | cutoff | N | old | new | speedup |
|---|--------|---|-----|-----|---------|
| 2 | 8  | 36  | 0.129s | 0.0005s | ~266x |
| 2 | 10 | 55  | 0.348s | 0.0009s | ~407x |
| 3 | 8  | 120 | 1.976s | 0.0051s | ~390x |

For tests I added `tests/_simulators/gaussian/test_density_matrix.py`, which checks the recurrence against the old per-element hafnian across d=1,2,3, displaced and not, a few cutoffs, plus a lossy/mixed state, Hermiticity, the diagonal lining up with `fock_probabilities`, and the fallback path. The full gaussian suite still passes.

One thing worth calling out: I added `hermite.py` to the ignore list in `.codecov.yml`. coverage.py can't see inside numba-jitted code, which is why `fock.py`, `indices.py`, `hafnian/*` and friends are already ignored - this is the same situation.

Note: I used Claude to help with parts of the implementation, but I worked through the math (Section III of the paper), reviewed the code, and tested it thoroughly against the existing per-element path - happy to walk through any part or adjust anything on review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # Changelog
 
 
+## [Unreleased]
+
+### Changed
+
+- `GaussianState.density_matrix` is now computed with the modified multidimensional
+  Hermite recurrence (Section III. of [arXiv:2209.06069](https://arxiv.org/abs/2209.06069))
+  instead of an independent (loop) hafnian per matrix element, sharing the work
+  across the whole truncated Fock space. This yields a ~100-400x speedup on the
+  NumPy connector for typical mode counts and cutoffs, with no change in results.
+
+
 ## [7.2.1] - 2026-03-21
 
 ### Fixed
--- a/piquasso/_math/hermite.py
+++ b/piquasso/_math/hermite.py
@@ -1,0 +1,156 @@
+#
+# Copyright 2021-2026 Budapest Quantum Computing Group
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""
+Multidimensional Hermite recurrence for Gaussian Fock-space amplitudes.
+
+The Fock-space representation of a Gaussian object (pure state, mixed state,
+unitary or channel) is characterized by a triple :math:`(A, b, c)`, where
+:math:`A` is a complex symmetric matrix, :math:`b` a complex vector and
+:math:`c` a complex scalar, see Section III. of
+https://arxiv.org/abs/2209.06069.
+
+The (renormalized) amplitudes :math:`\bar{G}^A_\mathbf{k}(b)` satisfy the
+order-2 linear recurrence relation (Eq. (45) therein)
+
+.. math::
+    \bar{G}_{\mathbf{k} + 1_i} = \frac{1}{\sqrt{k_i + 1}} \left (
+        b_i \bar{G}_\mathbf{k}
+        + \sum_j \sqrt{k_j} A_{ij} \bar{G}_{\mathbf{k} - 1_j}
+    \right ),
+    \qquad \bar{G}_\mathbf{0} = c,
+
+which generates *all* amplitudes in a single shared sweep instead of evaluating
+each Fock-space matrix element with an independent (loop) hafnian.
+
+For a mixed state on :math:`d` modes the multi-index :math:`\mathbf{k}` is
+:math:`2d`-dimensional and splits as ``ket`` :math:`\oplus` ``bra``, so the
+density matrix element :math:`\langle \mathrm{bra} | \rho | \mathrm{ket} \rangle`
+is exactly :math:`\bar{G}^A_{\mathrm{ket} \oplus \mathrm{bra}}(b)`.
+"""
+
+import numpy as np
+import numba as nb
+
+from piquasso._math.indices import get_index_in_fock_space
+
+
+@nb.njit(cache=True)
+def density_matrix_from_a_b_c(
+    A: np.ndarray,
+    b: np.ndarray,
+    c: complex,
+    basis: np.ndarray,
+) -> np.ndarray:
+    r"""Builds a Gaussian density matrix from its :math:`(A, b, c)` triple.
+
+    The element at position ``(i, j)`` equals
+    :math:`\langle \mathrm{basis}[i] | \rho | \mathrm{basis}[j] \rangle`, matching
+    ``DensityMatrixCalculation.get_density_matrix_element`` with ``bra = basis[i]``,
+    ``ket = basis[j]`` (i.e. ``reduce_on = ket ⊕ bra``).
+
+    Args:
+        A (numpy.ndarray): The :math:`2d \times 2d` complex symmetric matrix.
+        b (numpy.ndarray): The :math:`2d` complex vector (zeros for a
+            non-displaced state).
+        c (complex): The vacuum amplitude :math:`\langle 0 | \rho | 0 \rangle`.
+        basis (numpy.ndarray):
+            The occupation-number basis of the truncated Fock space, ordered by
+            increasing total particle number (as produced by
+            :func:`~piquasso._math.fock.get_fock_space_basis`).
+
+    Returns:
+        numpy.ndarray: The density matrix in the truncated Fock space.
+    """
+
+    d = A.shape[0] // 2
+    cardinality = basis.shape[0]
+
+    density_matrix = np.zeros(shape=(cardinality, cardinality), dtype=np.complex128)
+    density_matrix[0, 0] = c
+
+    for row in range(cardinality):
+        bra = basis[row]
+        for col in range(cardinality):
+            if row == 0 and col == 0:
+                continue
+
+            ket = basis[col]
+
+            if col > 0:
+                # The combined index has a nonzero `ket` block: raise there.
+                pivot = 0
+                for index in range(d):
+                    if ket[index] > 0:
+                        pivot = index
+                        break
+
+                ket_lowered = ket.copy()
+                ket_lowered[pivot] -= 1
+                col_lowered = get_index_in_fock_space(ket_lowered)
+
+                value = b[pivot] * density_matrix[row, col_lowered]
+
+                for index in range(d):
+                    if ket_lowered[index] > 0:
+                        ket_lowered2 = ket_lowered.copy()
+                        ket_lowered2[index] -= 1
+                        value += (
+                            np.sqrt(ket_lowered[index])
+                            * A[pivot, index]
+                            * density_matrix[row, get_index_in_fock_space(ket_lowered2)]
+                        )
+
+                for index in range(d):
+                    if bra[index] > 0:
+                        bra_lowered = bra.copy()
+                        bra_lowered[index] -= 1
+                        value += (
+                            np.sqrt(bra[index])
+                            * A[pivot, d + index]
+                            * density_matrix[
+                                get_index_in_fock_space(bra_lowered), col_lowered
+                            ]
+                        )
+
+                density_matrix[row, col] = value / np.sqrt(ket[pivot])
+            else:
+                # `ket` block is zero (col == 0): raise in the `bra` block. The
+                # `ket`-block sum in the recurrence vanishes identically.
+                pivot = 0
+                for index in range(d):
+                    if bra[index] > 0:
+                        pivot = index
+                        break
+
+                bra_lowered = bra.copy()
+                bra_lowered[pivot] -= 1
+                row_lowered = get_index_in_fock_space(bra_lowered)
+
+                value = b[d + pivot] * density_matrix[row_lowered, col]
+
+                for index in range(d):
+                    if bra_lowered[index] > 0:
+                        bra_lowered2 = bra_lowered.copy()
+                        bra_lowered2[index] -= 1
+                        value += (
+                            np.sqrt(bra_lowered[index])
+                            * A[d + pivot, d + index]
+                            * density_matrix[get_index_in_fock_space(bra_lowered2), col]
+                        )
+
+                density_matrix[row, col] = value / np.sqrt(bra[pivot])
+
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 # Changelog
 
 
+## [Unreleased]
```
Kolarovszki, line 4: This is a very good idea!

piquasso/_math/hermite.py
```diff
@@ -0,0 +1,156 @@
...
+        numpy.ndarray: The density matrix in the truncated Fock space.
+    """
+
+    d = A.shape[0] // 2
+    cardinality = basis.shape[0]
+
+    density_matrix = np.zeros(shape=(cardinality, cardinality), dtype=np.complex128)
+    density_matrix[0, 0] = c
+
+    for row in range(cardinality):
```
Kolarovszki, line 170: The code looks good, I need to clarify a few things for myself first. Until that, do you think it is possible to refactor it so that the code is less complex? I think it is 5 deep at this point. Can you perhaps reduce it to 2-3 by extracting some functions?

piquasso/_simulators/gaussian/probabilities.py
```diff
@@ -46,6 +48,13 @@ def get_density_matrix_element(self, bra: np.ndarray, ket: np.ndarray) -> float:
         )
 
     def get_density_matrix(self, occupation_numbers: np.ndarray) -> np.ndarray:
+        density_matrix = self.connector.density_matrix_from_representation(
+            self._A, self._b, self._normalization, occupation_numbers
+        )
+
+        if density_matrix is not None:
```
Kolarovszki: But then perhaps it would be nicer to have the original implementation put into `_BuiltinConnector` I suppose. I don't think it is difficult as `hafnian` and `loop_hafnian` are already available in a connector class. Then the fallback implementation would perhaps be differentiable immediately.

I'm not attached to `DensityMatrixCalculation` or its subclasses at all.

piquasso/api/connector.py
```diff
@@ -274,6 +274,27 @@ def calculate_interferometer_on_fock_space(self, interferometer, helper_indices)
...
+            b: The :math:`2d` complex vector (zeros for a non-displaced state).
+            c: The vacuum amplitude :math:`\\langle 0 | \\rho | 0 \\rangle`.
+            basis: The occupation-number basis of the truncated Fock space.
+
+        Returns:
+            The density matrix, or ``None`` when the connector provides no
+            accelerated implementation (the caller then falls back to the
+            generic per-element hafnian evaluation).
+        """
+        return None
```
Kolarovszki: I don't think this is a good practice. At this level, it is very unclear what `None` does, it only gets a meaning in `GaussianSimulator`. I would make this an abstract method, and have the fallback implementation in `_BuiltinSimulator` probably.

tests/_simulators/gaussian/test_density_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+import numpy as np
+
+import piquasso as pq
+
+from piquasso._math.fock import get_fock_space_basis
+
+
+def _reference_density_matrix(state):
+    """The original per-element (loop) hafnian evaluation, used as a reference."""
+    calculation = state._get_density_matrix_calculation()
```
Kolarovszki: Oh, I wouldn't use this private method. Can you please not rely on private stuff?

tests/_simulators/gaussian/test_density_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+        pq.Q(0) | pq.Attenuator(theta=0.3)
+
+    state = _execute(program, d, cutoff=7)
+
+    density_matrix = state.density_matrix
+
+    assert np.allclose(density_matrix, _reference_density_matrix(state))
+
+
+def test_connector_without_accelerated_path_returns_none():
```
Kolarovszki: I don't think that direct testing of connector methods are needed, and I would test the density matrix through `GaussianSimulator` instead. Users are not encouraged to use connector methods, and it is not considered an API breaking when these are refactored.

Kolarovszki (changes requested):
Overall, I like this patch! I am very happy that you have worked through the math.

CHANGELOG.md
JaipreethTiruvaipati: thanks...

piquasso/_math/hermite.py
JaipreethTiruvaipati: Good call. I pulled the recurrence apart into three helpers - _first_nonzero_mode, _entry_raising_ket and _entry_raising_bra. The main loop is now just for row / for col / if, and each helper is for / if, so nothing goes deeper than 3. Behaviour is unchanged (still the Eq. 45 recurrence), it just reads a lot better now.

piquasso/_simulators/gaussian/probabilities.py
JaipreethTiruvaipati: Done - moved the generic implementation into BuiltinConnector. get_density_matrix is now just a one-liner that delegates to the connector. The fallback uses loop_hafnian uniformly (with b = 0 it's exactly the plain hafnian), so it ends up differentiable wherever loop_hafnian is, like you suggested. I left DensityMatrixCalculation in place for now since it still builds the (A, b, c) triple and serves fock_probabilities, but happy to take it apart further if you'd prefer.

piquasso/api/connector.py
JaipreethTiruvaipati: Agreed, the None was confusing. It's an @abc.abstractmethod now, with the actual fallback living in BuiltinConnector (I read _BuiltinSimulator as BuiltinConnector, since hafnian/loop_hafnian are on the connector). NumpyConnector overrides it with the fast Hermite version, same pattern as calculate_interferometer_on_fock_space.

tests/_simulators/gaussian/test_density_matrix.py
JaipreethTiruvaipati: Fixed - dropped the private method entirely. Correctness is now pinned by comparing the NumPy (Hermite recurrence) and JAX (per-element fallback) density matrices through GaussianSimulator, plus an analytic thermal-state check. No internals touched.

tests/_simulators/gaussian/test_density_matrix.py
JaipreethTiruvaipati: Makes sense, removed it. Everything goes through GaussianSimulator now.

JaipreethTiruvaipati:
One extra thing worth flagging: to actually run the BuiltinConnector fallback on JAX (and get the "differentiable immediately" part), I had to fix a small crash in the JAX loop_hafnian - it was doing A[0, 0] = 1.0 on an immutable JAX array, now A.at[0, 0].set(1.0). Without it the fallback can't run on a non-NumPy connector at all, so it's also what lets the cross-connector test cover it. One known gap: single-mode squeezed vacuum still hits a separate numerical issue in the JAX eig-based placeholder, so I left that one case out of the cross-connector test. Glad to split the JAX fix into its own PR if you'd rather keep this one focused.

Kolarovszki:
> One known gap: single-mode squeezed vacuum still hits a separate numerical issue in the JAX eig-based placeholder

This is a bit unclear for me, can you clarify please what is the numerical issue? Is there a problem with the custom `eig` implementation from `piquasso/_math/jax/utils.py`?

tests/_simulators/gaussian/test_density_matrix.py
```diff
@@ -0,0 +1,124 @@
...
+        .state
+    )
+
+    assert np.allclose(
+        np.asarray(numpy_state.density_matrix),
+        np.asarray(jax_state.density_matrix),
+    )
+
+
+def test_thermal_density_matrix_matches_analytic():
```
Kolarovszki, line 77: Yes, such comparison to analytics is very reassuring.

Optional:
Can you check for single-mode squeezed and coherent states as well? The density operators are pretty easy also in these cases.

Kolarovszki (approved):

Kolarovszki:
Overall, I think this is a pretty good PR! Thank you for your contribution!

JaipreethTiruvaipati:
> > One known gap: single-mode squeezed vacuum still hits a separate numerical issue in the JAX eig-based placeholder
> 
> This is a bit unclear for me, can you clarify please what is the numerical issue? Is there a problem with the custom `eig` implementation from `piquasso/_math/jax/utils.py`?

Good question. It's not eig itself, it returns the right eigenvalues (both 0) there. The NaN is one step later in _loop_hafnian / _get_loop_polynom_coefficients, which does inv(O) on the eigenvector matrix. For single-mode squeezed vacuum the breaking elements are the odd-photon off-diagonal ones (e.g. ⟨0|ρ|1⟩), which are analytically zero. The reduced 1x1 zero, after the odd→even padding, is defective/non-diagonalizable (degenerate 0 eigenvalue), so O is effectively singular (cond ≈ 1e292) and inv(O) goes to Inf → NaN. So it's the eig-inverse step in the JAX loop-hafnian placeholder choking on a non-diagonalizable matrix, not eig or the density-matrix code. NumPy avoids it (power-trace/LaBudde, no eigenvector inverse) and the elements are zero anyway. Can open a separate issue for it.

tests/_simulators/gaussian/test_density_matrix.py
JaipreethTiruvaipati: Good idea, added single-mode squeezed-vacuum and coherent-state checks against their analytic density matrices. Both match.

JaipreethTiruvaipati:
> Overall, I think this is a pretty good PR! Thank you for your contribution!

Thanks a lot, the review made the patch noticeably cleaner. 🙏

JaipreethTiruvaipati:
Quick note on the red check - it's not the tests. On 3.14 the full suite passed (1443 passed, 139 skipped); the only failing step is Upload coverage to Codecov, and the other Python versions just show cancelled from the matrix fail-fast.

  The Codecov step errors with:
  ==> Fork detected, tokenless uploading used
  Could not pull latest version information: ... not valid JSON
  Report creating failed: {"message":"Token required because branch is protected"}
  So it's the tokenless fork upload against the protected base branch (with codecov.io also returning HTML instead of  JSON, which looks transient). The earlier commit 7888f3e uploaded fine at 94.97% with the same config, so nothing in the diff changed this.
  
  I re-ran the failed job to retry. If it sticks, it's the fork + protected-branch + CODECOV_TOKEN situation rather than the code - whatever's easiest on your end (re-run, or merge through it since tests are green).

Kolarovszki:
> > > One known gap: single-mode squeezed vacuum still hits a separate numerical issue in the JAX eig-based placeholder
> > 
> > This is a bit unclear for me, can you clarify please what is the numerical issue? Is there a problem with the custom `eig` implementation from `piquasso/_math/jax/utils.py`?
> 
> Good question. It's not eig itself, it returns the right eigenvalues (both 0) there. The NaN is one step later in _loop_hafnian / _get_loop_polynom_coefficients, which does inv(O) on the eigenvector matrix. For single-mode squeezed vacuum the breaking elements are the odd-photon off-diagonal ones (e.g. ⟨0|ρ|1⟩), which are analytically zero. The reduced 1x1 zero, after the odd→even padding, is defective/non-diagonalizable (degenerate 0 eigenvalue), so O is effectively singular (cond ≈ 1e292) and inv(O) goes to Inf → NaN. So it's the eig-inverse step in the JAX loop-hafnian placeholder choking on a non-diagonalizable matrix, not eig or the density-matrix code. NumPy avoids it (power-trace/LaBudde, no eigenvector inverse) and the elements are zero anyway. Can open a separate issue for it.

Oh, don't stress yourself, it is not part of the PR I think. I will handle this, but thank you!

Kolarovszki:
>   Quick note on the red check - it's not the tests. On 3.14 the full suite passed (1443 passed, 139 skipped); the only failing step is Upload coverage to Codecov, and the other Python versions just show cancelled from the matrix fail-fast.
> 
>   The Codecov step errors with:
>   ==> Fork detected, tokenless uploading used
>   Could not pull latest version information: ... not valid JSON
>   Report creating failed: {"message":"Token required because branch is protected"}
>   So it's the tokenless fork upload against the protected base branch (with codecov.io also returning HTML instead of  JSON, which looks transient). The earlier commit 7888f3e uploaded fine at 94.97% with the same config, so nothing in the diff changed this.
>   
>   I re-ran the failed job to retry. If it sticks, it's the fork + protected-branch + CODECOV_TOKEN situation rather than the code - whatever's easiest on your end (re-run, or merge through it since tests are green).
> 

Codecov is always problematic for some reason :/ I will look into this soon!

Kolarovszki:
@JaipreethTiruvaipati Can you comment on the issue (so that I can assign you)?

JaipreethTiruvaipati:
Sure, i will do that... Thanks @Kolarovszki

PR #607, opened 2026-06-16
Fix ParticleNumberMeasurement to measure only specified modes on SamplingSimulator (#499)

### LINKED ISSUE #499: Always all modes are measured for ``ParticleNumberMeasurement`` on ``SamplingSimulator``

mk0dz:
Closes #499 
## The bug
On `SamplingSimulator`, `ParticleNumberMeasurement` ignored the modes in `pq.Q(...)`
and always returned all modes:
```python
pq.Q(0, 1) | pq.ParticleNumberMeasurement()      # d=5, StateVector([1,1,1,0,0])
# SamplingSimulator  -> (1, 1, 1, 0, 0)   (wrong)
# PureFockSimulator  -> (1, 1)            (correct)
```
In `sampling/simulation_steps.py::particle_number_measurement`, the
generalized-Clifford sampler produces a full-mode sample, which was binned
directly — `instruction.modes` was never used.

## The fix
Project each full-mode sample onto the measured modes before binning. For a
terminal particle-number measurement, the marginal over a subset of modes is
exactly "measure all, keep the requested columns" — and projecting in the given
order reproduces `PureFockSimulator`, which does `state.reduced(instruction.modes)`.

```python
measured_modes = instruction.modes
samples = [tuple(sample[mode] for mode in measured_modes) for sample in samples]
```
No new dependency, no analytic-marginal machinery, no performance cliff.

## Validation (local, editable build, Python 3.13)
Cross-checked against `PureFockSimulator` (the correct reference):

| input / modes | SamplingSimulator | PureFockSimulator |
|---|---|---|
| `[1,1,1,0,0]`, modes `(0,1)` | `(1,1)` | `(1,1)` ✓ |
| `[1,1,1,0,0]`, modes `(0,1,2,3,4)` | `(1,1,1,0,0)` | same ✓ |
| `[1,0,1,0,1]`, modes `(2,3)` | `(1,0)` | same ✓ |
| `[2,0,1,0]`, modes `(3,0)` (reordered) | `(0,2)` | same ✓ |
| `[1,0]` + 50/50 beamsplitter, mode `(0,)` | `{(0):0.50,(1):0.50}` | `{(0):0.48,(1):0.52}` ✓ (within noise) |

## Files changed (PR contents)
```
 piquasso/_simulators/sampling/simulation_steps.py   (+8: project samples to measured modes)
 tests/_simulators/sampling/test_measurements.py     (+31: 2 regression tests)
 CHANGELOG.md                                         (+ Fixed entry under [Unreleased])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `ParticleNumberMeasurement` on `SamplingSimulator` now reports only the modes
+  specified in `pq.Q(...)`, instead of always measuring all modes, matching
+  `PureFockSimulator`'s behaviour
+  ([#499](https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues/499)).
+
 ### Changed
 
 - `GaussianState.density_matrix` is now computed with the modified multidimensional
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@
   `PureFockSimulator`'s behaviour
   ([#499](https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues/499)).
 
+### Added
+
+- An efficient marginal sampling algorithm for `ParticleNumberMeasurement` on
+  `SamplingSimulator` when only a few modes are measured: instead of sampling all
+  modes and discarding, the marginal distribution over the requested modes is
+  computed directly via a Laguerre-polynomial generating function (valid for
+  collided inputs). This is hundreds of times faster than full-sample-then-discard
+  for small mode counts; a FLOP heuristic falls back to discarding for large ones
+  ([#499](https://github.com/Budapest-Quantum-Computing-Group/piquasso/issues/499)).
+
 ### Changed
 
 - `GaussianState.density_matrix` is now computed with the modified multidimensional
--- a/benchmark/sampling_marginal_benchmark.py
+++ b/benchmark/sampling_marginal_benchmark.py
@@ -1,0 +1,69 @@
+#
+# Copyright 2021-2026 Budapest Quantum Computing Group
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Benchmark of ParticleNumberMeasurement on a strict subset of modes (issue #499):
+the efficient marginal algorithm vs. full-sampling-then-discarding.
+
+This justifies the switch implemented in
+``piquasso._simulators.sampling.marginal.prefer_marginal_sampling``: for small
+numbers of measured modes ``k`` the marginal algorithm is orders of magnitude
+faster, while for large ``k`` the discard approach wins (and the heuristic falls
+back to it).
+
+Run with: ``python benchmark/sampling_marginal_benchmark.py``
+"""
+
+import time
+
+import numpy as np
+from scipy.stats import unitary_group
+
+import piquasso as pq
+import piquasso._simulators.sampling.simulation_steps as steps
+
+
+def _time_execution(d, occupation, modes, shots, force_discard):
+    U = unitary_group.rvs(d, random_state=3)
+
+    with pq.Program() as program:
+        pq.Q() | pq.StateVector(occupation)
+        pq.Q(*range(d)) | pq.Interferometer(U)
+        pq.Q(*modes) | pq.ParticleNumberMeasurement()
+
+    original = steps.prefer_marginal_sampling
+    if force_discard:
+        steps.prefer_marginal_sampling = lambda *args, **kwargs: False
+    try:
+        start = time.perf_counter()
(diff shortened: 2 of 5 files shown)


Kolarovszki (changes requested):
This is very insufficient. If the specified modes is small, there should be an efficient sampling algorithm, see discussion under the issue.

mk0dz:
Thanks for the clear feedback! implemented the efficient marginal algorithm you outlined in the issue.
  
  `piquasso/_simulators/sampling/marginal.py` builds the Laguerre generating
  function P(z,w) = Π_a L_{r_a}(-f_a(z)·conj_coeff(f_a(w))), extracts the diagonal coefficients P_{α,α}, and forms Pr(Y=y) via your formula. `particle_number_ measurement` uses it for a strict subset of measured modes, and as per your note about big k - falls back to full-sampling-then-discard via a FLOP heuristic
  (`prefer_marginal_sampling`).
  
  Verification:
  - The marginal probabilities match a brute-force boson-sampling marginalization
    to 5e-16, including collided inputs.
  - They match `PureFockSimulator(...).state.reduced(modes).fock_probabilities_map`
    to 2.6e-16.
  - End-to-end (interferometer + 40k shots), the SamplingSimulator distribution
    agrees with PureFockSimulator (TVD ≈ 0.005, sampling noise).
  
  Benchmark (`benchmark/sampling_marginal_benchmark.py`, 2000 shots):
  
  | d | k | efficient | discard | speedup |
  |---|---|---|---|---|
  | 6 | 1 | 0.80 ms | 751 ms | 937× |
  | 6 | 2 | 1.05 ms | 751 ms | 716× |
  | 6 | 3 | 3.03 ms | 758 ms | 250× |
  | 8 | 2 | 1.03 ms | 791 ms | 765× |
  
  The cost growing with k is exactly why the heuristic flips back to discarding for
  large k. Added tests + a CHANGELOG entry. Happy to tune the crossover or switch to
  mode-by-mode sampling if you'd prefer.

Kolarovszki (changes requested):
Are you sure about the cost of this marginal algorithm? (I'm talking about `prefer_marginal_sampling`) So there are n^{2k} coefficients to be computed (actually, there are significantly less), but you use the naive algorithm for the polynomial multiplication, which is not the best possible algorithm for this task.

mk0dz:
Checked the cost empirically  and  you're right, it was wrong. Measured the actual size of the (z,w) coefficient table the build produces (collision-free inputs):
  
  | n, k | tracked coeffs | n^{2k} | C(n+k,k)^2 |
  |------|----------------|--------|------------|
  | 6, 2 | 140            | 1296   | 784        |
  | 8, 2 | 285            | 4096   | 2025       |
  | 6, 3 | 1596           | 46656  | 7056       |
  
The |α| = |β| (equal total degree) constraint prunes the table well below n^{2k}, and the diagonal we actually need is only C(n+k, k). My old `(n+1)^{2k+1}` estimate was far too pessimistic, so the heuristic switched to discarding earlier than it should. Replaced it: marginal cost ≈ C(n+k, k)^2 (the dominant cost is the contraction forming Pr(Y=y) from the C(n+k, k) diagonal coefficients), discard ≈ shots · n · 2^n.
  
On the polynomial multiplication — it is the naive convolution. Each factor L_{r_a}(g_a) goes in as a low-degree term (for r_a = 1 it's just 1 − f_a(z)·conj(f_a(w)), O(k^2) terms), so each step is roughly linear in the current table size, but I'm not claiming it's optimal, especially for collided inputs where the factor degree grows. Did you have a particular approach in mind (FFT-based convolution, or a DP that contracts straight to the diagonal)? I'd rather implement the one you'd prefer than guess.

Kolarovszki:
@mk0dz As efficient as reasonably achievable. See discussions at other PRs referencing the issue.

mk0dz:
Read through the #590 discussion. A few things on where this PR stands:
  
  The implementation already keeps only the bihomogeneous coefficients — it stores
  P_{α,β} in a sparse dict and, because every factor F_a is bilinear, only |α|=|β|
  terms are ever created (verified: no |α|≠|β| keys appear). So it's the sparse,
  zero-avoiding, truncated representation rather than a dense N^k array. 
  
  The thing I wanted to flag: this exact-arithmetic route stays numerically stable
  where the roots-of-unity/grid approach in #590 diverged. Measured sum of the
  marginal probabilities (k=2): 
  
    n=10 -> 1.000000000000
    n=15 -> 1.000000000000
    n=18 -> 1.000000000000   (the grid approach reportedly hit ~2239 here)
    n=20 -> 1.000000000000   
    
  Timing (k=2): n=12 ~14ms, n=18 ~63ms, n=20 ~94ms; k=1 stays <1ms. Polynomial in n
  for fixed k, as expected. 
  
  Where it's not optimal: the multiplication is sparse-dict convolution, not FFT.
  For collision-free inputs each step multiplies by a small low-degree F_a, so it's
  near-linear in the current table; the FFT-based bihomogeneous multiplication you
  described would mainly help high-collision inputs (large r_a) and shave the
  constant. I can pursue that, but it's the part you flagged as hard, and it's what
  introduced the instability in the grid attempt — so I wanted to check: is the
  stable sparse version acceptable here, or do you specifically want the FFT
  multiplication for this PR?

Kolarovszki:
@mk0dz Sorry, I thought I sent this comment earlier -- I had bad connection.

I would like to see an algorithm which is as efficient as reasonably achievable, which is not necessarily a convenient one. In particular, the final algorithm might be a bit ugly, I expected something much simpler initially. See comments at other PRs.

Kolarovszki:
Closed due to inactivity.

